PR #11, opened 2025-10-31
Allow reading config from files

samshadwell:
### Summary

Thanks for making this MCP server!

I'm deploying it inside a Docker container, and the recommended way to pass secrets there is file-based. From [Docker's documentation on the subject](https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images):

> If you develop a container that can be deployed as a service and requires sensitive data, such as a credential, as an environment variable, consider adapting your image to take advantage of Docker secrets. One way to do this is to ensure that each parameter you pass to the image when creating the container can also be read from a file.

This PR does this!

### Testing

Tested out locally, both the error and non-error cases. Also tried password files that included trailing whitespace (like from `echo`) and without (like from `echo -n` or `printf`).

Success:
Screenshot 2025-10-31 at 10 04 23 AM

Error:
Screenshot 2025-10-31 at 10 02 58 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,17 @@ uv sync
 
 ## Running the Server
 
+### Configuration
+
+Your Garmin Connect credentials are read from environment variables:
+
+- `GARMIN_EMAIL`: Your Garmin Connect email address
+- `GARMIN_EMAIL_FILE`: Path to a file containing your Garmin Connect email address
+- `GARMIN_PASSWORD`: Your Garmin Connect password
+- `GARMIN_PASSWORD_FILE`: Path to a file containing your Garmin Connect password
+
+File-based secrets are useful in certain environments, such as inside a Docker container. Note that you cannot set both `GARMIN_EMAIL` and `GARMIN_EMAIL_FILE`, similarly you cannot set both `GARMIN_PASSWORD` and `GARMIN_PASSWORD_FILE`.
+
 ### With Claude Desktop
 
 1. Create a configuration in Claude Desktop:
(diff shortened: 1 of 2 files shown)


Taxuspt (approved):

msillingworth:
Implemented in 53026ec.

Autonomous Mode's daily orchestration is live in shadow mode: every accepted proposal is persisted (autonomous_store.py) and logged, but upload_workout is never called.

**Delivered:**
- Two-layer decision model wired end-to-end: Guardrails (guardrails.py) always retain final veto over Workout Composition's (autonomous_llm.py) LLM judgment.
- Stability Gradient: days 1-2 are frozen (only a Guardrail override can change them, never LLM re-composition); days 3-7 are freely refined each run.
- Each goal's open days are composed in a **single batched Claude call**, not one call per day — verified live (1 call for a 7-day window).
- `evaluation_summary` in the Workout Composition prompt is now a real `evaluation.evaluate_goal()` / `route_effort.evaluate_route_effort_goal()` result.
- Fail-safe throughout: a systemic error (Garmin auth, context gathering) results in zero action and a "failed" decision-log entry; a per-goal error (e.g. Claude API failure) is isolated to that goal only; a malformed/unparseable LLM response blocks rather than guesses.

**Two live-verification bugs found and fixed during implementation** (see ADR-0004's correction note and PRD.md):
1. `readiness_age_hours` was computed against midnight-of-today instead of the actual current time, making same-day morning readings appear to be from the future.
2. `check_cross_sport_load_cap` compared `dailyAcuteChronicWorkloadRatio` (~0-2 scale) against `min/maxTrainingLoadChronic` bounds that are actually in the hundreds — a dimensional mismatch that would have silently blocked every single Autonomous Mode proposal forever. Fixed to compare `dailyTrainingLoadChronic` instead.

Full suite (624 tests) + ruff clean on all touched files. Verified against the live account in both blocked (no recent activity history) and accepted (real cycling history) paths.

Issue #63, opened 2026-03-30
Garth Deprecated and Garmin Auth no longer working


brunosantos:
Hi,
Garmin has [made some changes around Auth](https://forums.garmin.com/apps-software/mobile-apps-web/f/garmin-connect-web/433892/tls-fingerprinting-blocks-third-party-clients) and broke Garth and [GarminConnect](https://github.com/cyberjunky/python-garminconnect/issues/332). As a result [matin ](https://github.com/matin)has deprecated the Garth project (https://github.com/matin/garth/discussions/222). Does anyone have an alternative Garmin Auth lib that can still get the garmin connect access tokens?

nrvim:
For anyone dealing with the 401 errors and broken auth — I built **[garmin-givemydata](https://github.com/nrvim/garmin-givemydata)**, a completely different approach that bypasses the OAuth/garth dependency entirely.

  ### How it's different

Instead of fighting Garmin's OAuth flow (which keeps breaking), it uses Playwright browser automation — a real Chrome browser that logs in exactly like you would. Cloudflare can't
  tell it apart from a normal user because it IS a normal browser.

  - First run opens Chrome for login + MFA, saves the session. All in your local database.
  - All subsequent runs are fully headless — session persists ~1 year
  - Works with accounts that get 401 errors on the OAuth endpoint (since it doesn't use OAuth at all)

  ### What it collects (47 SQLite tables)

  | Category | Data |
  |----------|------|
  | **Daily health** | HR, HRV, stress, sleep stages, body battery, SpO2, respiration, steps, floors, calories, hydration, fitness age |
  | **Training** | Training readiness with factor breakdowns, endurance score, hill score, race predictions (5K/10K/half/marathon), VO2max trends |
  | **Body** | Weight, BMI, body fat, body water, bone mass, muscle mass |
  | **Activities** | 45+ fields per activity, plus splits, HR zones, weather, exercise sets, running dynamics |
  | **Profile** | Earned badges, personal records, gear, devices |
  | **Files** | Original FIT file downloads for every activity |

  ### Smart sync

  First run fetches your full history (back to when you created your Garmin account). After that, incremental syncs take seconds — only fetches what's new.

  ### Built-in MCP server

  Works with Claude Code, Claude Desktop, or any MCP client. 6 tools including free-form SQL queries across all 47 tables. Ask things like *"How has my HRV changed over 6 months?"* or *"Give me a full sports medicine health check"* and
  it queries your local database directly.

[...]

PR #70, opened 2026-04-02
Fix: can't login - getting 429

brunosantos:
https://github.com/Taxuspt/garmin_mcp/issues/63

Garmin has [made some changes around Auth](https://forums.garmin.com/apps-software/mobile-apps-web/f/garmin-connect-web/433892/tls-fingerprinting-blocks-third-party-clients) and broke Garth and https://github.com/cyberjunky/python-garminconnect/issues/332. As a result [matin ](https://github.com/matin)has deprecated the Garth project (https://github.com/matin/garth/discussions/222). This change follows [the suggestion made here](https://github.com/matin/garth/discussions/222#discussioncomment-16418591) and seems to work.

Update User-Agent header for Garmin session.
workaround Cloudflares block with user agent spoofing

### DIFF
--- a/src/garmin_mcp/auth_cli.py
+++ b/src/garmin_mcp/auth_cli.py
@@ -125,6 +125,14 @@ def authenticate(token_path: str, token_base64_path: str, force_reauth: bool = F
 
     try:
         garmin = Garmin(email=email, password=password, is_cn=is_cn, prompt_mfa=get_mfa)
+        garmin.garth.sess.headers.update({
+            "User-Agent": (
+                "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
+                "AppleWebKit/537.36 (KHTML, like Gecko) "
+                "Chrome/131.0.0.0 Safari/537.36"
+            )
+        })
+    
         garmin.login()
 
         # Save tokens to directory


mknightz:
The auth fix itself (user-agent spoofing + garth token exchange patch) is reasonable as a short-term workaround, but there are two concerns:

**1. Scope — this PR deletes most of the tools.** The diff shows 6,482 deletions across `challenges.py`, `nutrition.py`, `data_management.py`, `devices.py`, `gear_management.py`, `health_wellness.py`, `training.py`, `user_profile.py`, `weight_management.py`, `womens_health.py`, and their tests. This takes the project from ~96 tools to ~22. The auth fix is +27 lines — the rest is unrelated removal. These should be separate PRs.

**2. User-agent spoofing is fragile.** Garmin's block is at the TLS fingerprinting level (Cloudflare), not just user-agent headers. This may work today but is likely to break again. `garminconnect==0.3.2` has solved this properly by switching to `curl_cffi`, which can present browser-like TLS fingerprints. PR #77 takes this approach.

I'd suggest scoping this to just the auth changes and considering whether #77's upstream library approach is more sustainable long-term.

brunosantos:
> The auth fix itself (user-agent spoofing + garth token exchange patch) is reasonable as a short-term workaround, but there are two concerns:
> 
> **1. Scope — this PR deletes most of the tools.** The diff shows 6,482 deletions across `challenges.py`, `nutrition.py`, `data_management.py`, `devices.py`, `gear_management.py`, `health_wellness.py`, `training.py`, `user_profile.py`, `weight_management.py`, `womens_health.py`, and their tests. This takes the project from ~96 tools to ~22. The auth fix is +27 lines — the rest is unrelated removal. These should be separate PRs.
> 
> **2. User-agent spoofing is fragile.** Garmin's block is at the TLS fingerprinting level (Cloudflare), not just user-agent headers. This may work today but is likely to break again. `garminconnect==0.3.2` has solved this properly by switching to `curl_cffi`, which can present browser-like TLS fingerprints. PR #77 takes this approach.
> 
> I'd suggest scoping this to just the auth changes and considering whether #77's upstream library approach is more sustainable long-term.

Totally agree. The tool removal was not intended for this PR. Let me fix that later today.

brunosantos:
updated PR to only have the single user-agent change.
I think this is the quickest way to get a tactical solution out while PR https://github.com/Taxuspt/garmin_mcp/pull/77 is in review

Muskos:
Thanks, this changes work for me

Taxuspt:
Closing — User-Agent spoofing doesn't fix the underlying TLS fingerprinting issue. The root fix landed in #77 (upgrade to garminconnect 0.3.2 with curl_cffi).

PR #133, opened 2026-06-09
Validate workout target type mappings

### LINKED ISSUE #129: heart.rate target type on workout steps stored as pace.zone instead
When creating a workout step with a target type of heart.rate and lower/upper HR values, the target is stored and returned as pace.zone instead. The values are preserved correctly but the target type is wrong.

**To reproduce**

Create a running workout step using:

"targetType": {
  "workoutTargetTypeId": 6,
  "workoutTargetTypeKey": "heart.rate"
},
"targetValueOne": 143,
"targetValueTwo": 157

Fetch the workout back via get_workout_by_id -- the returned step shows target_type: pace.zone with the same values 143 and 157.

**Expected behaviour**

The step targets a custom HR range of 143-157 bpm and the watch alerts when HR goes outside that range.

**Actual behaviour**

The target type is stored as pace.zone. The watch interprets the values as pace targets, not HR targets.

**Workaround**

Use heart.rate.zone with a zone number instead of a custom HR range. This limits precision but stores correctly.

**Related issue**

This appears to be the same class of mapping bug as the heart.rate end condition issue, where heart.rate is mapped to calories on recovery steps.

pshlos46:
## Summary

Fixes #129.

This PR adds pre-upload validation for Garmin workout `targetType` ID/key pairs. Garmin treats `workoutTargetTypeId` as authoritative, so a payload such as:

```json
{"workoutTargetTypeId": 6, "workoutTargetTypeKey": "heart.rate"}
```

is stored by Garmin as `pace.zone`, because ID `6` is the canonical `pace.zone` target type. The validator now rejects known ID/key mismatches before upload instead of allowing Garmin to silently create the wrong workout target.

## Reproduction

I reproduced the issue against Garmin Connect before changing the code:

- Uploaded a running workout step with `workoutTargetTypeId: 6` and `workoutTargetTypeKey: "heart.rate"`, with `targetValueOne: 143` and `targetValueTwo: 157`.
- Fetched the workout back via `get_workout_by_id`.
- Garmin returned `workoutTargetTypeId: 6` with `workoutTargetTypeKey: "pace.zone"`; the numeric values were preserved as `143.0` and `157.0`.
- Uploaded a control workout using `workoutTargetTypeId: 4` and `workoutTargetTypeKey: "heart.rate.zone"` with the same `targetValueOne` / `targetValueTwo` values.
- Garmin returned `heart.rate.zone` and preserved the custom HR range.
- Both temporary repro workouts were deleted successfully.

## Technical choices

I chose strict validation rather than silently rewriting mismatched payloads because the malformed payload is ambiguous. If a caller sends ID `6` and a heart-rate key, we cannot safely know whether they intended pace or heart rate. Rewriting the request could hide a client-side bug and still produce a workout that behaves differently from what the caller requested.

This does not conflict with the existing HR-zone compatibility fix from #49. That fix handles a narrower, non-ambiguous case: `heart.rate.zone` with `targetValueOne` set to a small integer `1..5`, which is almost certainly a named zone number and is normalized to `zoneNumber`. Custom HR bpm ranges like `143..157` are still valid and are left unchanged. The new validation runs after that normalization, preserving the old behavior while rejecting true ID/key mismatches.

The canonical custom HR range form is:

```json
{
  "targetType": {
    "workoutTargetTypeId": 4,
    "workoutTargetTypeKey": "heart.rate.zone"
  },
  "targetValueOne": 143,
  "targetValueTwo": 157
}
```

Named Garmin HR zones continue to use the same target type with `zoneNumber` instead of `targetValueOne` / `targetValueTwo`.

## Changes

- Added known `targetType` canonical mappings and recursive pre-upload validation.
- Applied validation in `upload_workout`, `upload_workouts`, and inline `schedule_workouts` uploads.
- Kept the existing HR-zone normalization behavior intact.
- Documented raw `upload_workout` target type mapping and custom HR range usage in the README.
- Added integration coverage for invalid mismatch rejection, valid custom HR ranges, nested repeat groups, batch uploads, and inline scheduled uploads.

## Test plan

- [x] `uv run pytest tests/integration/test_workouts_tools.py` - 43 passed
- [x] `uv run pytest tests/e2e tests/integration tests/unit tests/test_mcp_debug.py` - 307 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,39 @@ Common end-condition IDs:
 | 10 | `reps` |
 | 11 | `training.peaks.tss` |
 
+### Raw `upload_workout` target types
+
+When building raw Garmin workout JSON, `targetType.workoutTargetTypeId` and
+`targetType.workoutTargetTypeKey` must use Garmin's canonical mapping. Garmin
+treats the numeric ID as authoritative: a mismatched payload such as
+`{"workoutTargetTypeId": 6, "workoutTargetTypeKey": "heart.rate"}` is stored as
+`pace.zone`, because ID `6` means `pace.zone`.
+
+For a custom heart-rate range, use target type ID `4` with `heart.rate.zone` and
+put the bpm range in `targetValueOne` / `targetValueTwo`:
+
+```json
+{
+  "targetType": {
+    "workoutTargetTypeId": 4,
+    "workoutTargetTypeKey": "heart.rate.zone"
+  },
+  "targetValueOne": 143,
+  "targetValueTwo": 157
+}
+```
+
+For a named Garmin HR zone, use the same target type with `zoneNumber` instead:
+
+```json
+{
+  "targetType": {
+    "workoutTargetTypeId": 4,
+    "workoutTargetTypeKey": "heart.rate.zone"
+  },
+  "zoneNumber": 3
+}
+```
 ## One-click Install (Claude Desktop)
 
 The easiest way to add this server to Claude Desktop is via the `.dxt` Desktop Extension file — no JSON editing required.
--- a/src/garmin_mcp/workouts.py
+++ b/src/garmin_mcp/workouts.py
@@ -61,7 +61,7 @@ def _fix_hr_zone_step(step: dict) -> None:
     Custom HR bpm ranges (e.g. targetValueOne=105, targetValueTwo=143) are left
     unchanged — these are legitimate custom heart rate targets in Garmin Connect.
     """
-    target_type = step.get('targetType', {})
+    target_type = step.get('targetType') or {}
     target_key = target_type.get('workoutTargetTypeKey', '')
 
     if target_key == 'heart.rate.zone' and 'zoneNumber' not in step:
--- a/src/garmin_mcp/workouts.py
+++ b/src/garmin_mcp/workouts.py
@@ -160,9 +160,9 @@ def _validate_end_condition_steps(workout_data: dict) -> None:
             _validate_end_condition_step(step, path)
 
 
-def _validate_target_type_step(step: dict, path: str) -> None:
-    """Reject targetType id/key pairs Garmin would silently reinterpret."""
-    target_type = step.get('targetType')
+def _validate_target_type_block(step: dict, path: str, target_field: str) -> None:
+    """Reject a target type id/key pair Garmin would silently reinterpret."""
+    target_type = step.get(target_field)
     if isinstance(target_type, dict):
         target_key = target_type.get('workoutTargetTypeKey')
         target_id = target_type.get('workoutTargetTypeId')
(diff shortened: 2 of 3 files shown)


pshlos46:
A note on mapping scope: this PR intentionally validates only known/verified `targetType` mappings (`1 -> no.target`, `4 -> heart.rate.zone`, `6 -> pace.zone`) instead of trying to enforce the full Garmin target-type enum.

The reason is compatibility. Garmin may support additional target types that are not represented in this repo's templates or fixtures yet, such as sport-specific targets or targets we have not captured from Garmin-created workouts. Rejecting every unknown ID now could block valid Garmin payloads just because our local mapping is incomplete.

So the chosen behavior is:

- reject known ID/key mismatches that Garmin would silently reinterpret, such as `6 + heart.rate` becoming `pace.zone`
- allow unknown IDs/keys for now, until they are verified and added to the known mapping
- make the mapping name/comment explicit in code (`KNOWN_TARGET_TYPE_IDS`) so reviewers do not read it as a complete Garmin enum

This keeps the fix strict where we have evidence, while avoiding an unnecessary backwards-compatibility break for target types we have not validated yet.

pshlos46:
Follow-up after review: I checked whether the same issue can happen on `secondaryTargetType` before changing the PR further.

This is now confirmed, not just theoretical. I uploaded temporary swim workouts because Garmin swim workouts can store pace prescriptions in `secondaryTargetType` while `targetType` is `null`.

Mismatch probe input:

```json
{
  "secondaryTargetType": {
    "workoutTargetTypeId": 6,
    "workoutTargetTypeKey": "heart.rate"
  },
  "secondaryTargetValueOne": 143,
  "secondaryTargetValueTwo": 157
}
```

Fetched back from Garmin:

```json
{
  "secondaryTargetType": {
    "displayOrder": 6,
    "workoutTargetTypeId": 6,
    "workoutTargetTypeKey": "pace.zone"
  },
  "secondaryTargetValueOne": 143.0,
  "secondaryTargetValueTwo": 157.0
}
```

A control upload with `6 + pace.zone` also came back as `pace.zone`, as expected. Both temporary workouts were deleted successfully.

I updated the PR so the same known mapping validation now applies to both `targetType` and `secondaryTargetType`, including nested repeat-group steps.

Additional tests added:

- flat `secondaryTargetType` mismatch rejection
- nested `secondaryTargetType` mismatch rejection

Updated test results:

- `uv run pytest tests/integration/test_workouts_tools.py` - 45 passed
- `uv run pytest tests/e2e tests/integration tests/unit tests/test_mcp_debug.py` - 309 passed

pshlos46:
Follow-up fix for the review finding around realistic swim secondary-target payloads:

Garmin swim workouts can use `targetType: null` with a populated `secondaryTargetType`. Before target validation runs, the existing HR-zone normalization path assumed `targetType` was always a dict, so `targetType: null` could raise before upload/validation.

Changes added:

- made `_fix_hr_zone_step` null-safe with `step.get("targetType") or {}`
- updated secondary-target mismatch tests to use `targetType: null`
- added a valid secondary pace target test with `targetType: null` to cover the real Garmin swim shape

Updated tests:

- `uv run pytest tests/integration/test_workouts_tools.py` - 46 passed
- `uv run pytest tests/e2e tests/integration tests/unit tests/test_mcp_debug.py` - 310 passed

Issue #132, opened 2026-06-08
Add support for unscheduling/removing workouts from Garmin calendar


skyf0l:
## Feature request

Add a tool to remove a scheduled workout from the Garmin Connect calendar without deleting the workout from the workout library.

## Problem

The MCP currently supports creating/uploading workouts and scheduling them to a calendar date, but I could not find a reliable way to remove or replace a workout that is already scheduled on the Garmin calendar.

Current related tools include:

* `schedule_workout`
* `schedule_week`
* `schedule_workouts`
* `get_scheduled_workouts`
* `delete_workout`
* `delete_workouts`

The missing piece is an explicit calendar-level delete/unschedule operation.

## Use case

Example flow:

1. Create workout A.
2. Schedule workout A on `2026-06-11`.
3. Later create workout B as an improved version.
4. Want to replace A with B on the same date.

Right now, `schedule_workout` / `schedule_week` appear to be idempotent and avoid duplicating workouts on an occupied date, which is good. But without an unschedule tool, the user has to manually remove the old workout from Garmin Connect before scheduling the new one.

### `unschedule_workout`

Remove one scheduled workout from a specific date.

Arguments:

```json
{
  "workout_id": 123456789,
  "calendar_date": "2026-06-11"
}
```

Expected behavior:

* Removes only the scheduled calendar entry.
* Does not delete the workout from the workout library.
* Returns success if the entry existed and was removed.
* Returns a clear no-op response if the workout was not scheduled on that date.

### `unschedule_workouts`

Batch version.

Arguments:

```json
{
  "schedules": [
    {
      "workout_id": 123456789,
      "calendar_date": "2026-06-11"
    },
    {
      "workout_id": 987654321,
      "calendar_date": "2026-06-14"
    }
  ]
}
```

## Nice-to-have: replace behavior

A higher-level `replace_scheduled_workout` tool would also be useful.

Arguments:

```json
{
  "calendar_date": "2026-06-11",
  "old_workout_id": 123456789,
  "new_workout_id": 987654321
}
```

Or:

```json
{
  "calendar_date": "2026-06-11",
  "new_workout_id": 987654321,
  "replace_existing": true
}
```

Expected behavior:

* If there is already a workout on that date, remove it and schedule the new one.
* If multiple workouts exist on that date, either fail with a clear error or require an explicit `old_workout_id`.
* Avoid duplicate calendar entries.

## Why this matters

[...]

PR #134, opened 2026-06-09
feat: add unschedule_workout and unschedule_workouts tools

valdugay:
## Summary

Partially addresses #132.

Adds the ability to remove a scheduled workout from the Garmin Connect
calendar, completing the schedule/unschedule pair. The workout template is
left untouched in the library — only the calendar entry is removed.

## Changes

1. **`get_scheduled_workouts` now exposes `scheduled_workout_id`** — the
   calendar-entry id from the GraphQL `workoutScheduleSummariesScalar`
   response. This is distinct from `workout_id` and is the id required to
   unschedule.
2. **`unschedule_workout(scheduled_workout_id)`** — removes a single calendar
   entry, leaving the workout template in the library intact.
3. **`unschedule_workouts(scheduled_workout_ids)`** — batch variant, same
   aggregate result shape as `delete_workouts`.

Both delegate to the high-level `garminconnect.unschedule_workout`, which
returns a dict (not a `Response`), so failure is signalled via exceptions —
the same pattern used by `delete_workout`.

## Scope relative to #132

This PR covers the core ask (an explicit calendar-level unschedule, single +
batch) but intentionally differs from the issue's proposed shape in one place,
and leaves the nice-to-have for a follow-up:

- **Identified by `scheduled_workout_id`, not `(workout_id, calendar_date)`.**
  The same workout can be scheduled multiple times (even on the same date), so
  the calendar-entry id is the unambiguous identifier. It is surfaced directly
  by `get_scheduled_workouts`, so the natural flow is *list → unschedule by id*,
  mirroring the existing `get_workouts` → `delete_workout` pattern. A
  `(workout_id, calendar_date)` convenience wrapper that resolves the id could
  be layered on top later if preferred.
- **`replace_scheduled_workout` is not included.** With `schedule_workout`
  (idempotent) and `unschedule_workout` now available, replace is composable;
  a dedicated higher-level tool can follow in a separate PR.

### Follow-ups (happy to do these in separate PRs)

- **`unschedule_workout_on_date(workout_id, calendar_date)`** — a convenience
  wrapper matching the signature proposed in #132. It would resolve the
  `scheduled_workout_id` (e.g. via the calendar lookup) and then delete, so
  callers that only have `workout_id` + date don't need a prior
  `get_scheduled_workouts` call. This is a nicer ergonomic for LLM-driven
  flows and I'm keen to add it.
- **`replace_scheduled_workout(calendar_date, new_workout_id, ...)`** — the
  higher-level replace from #132, built on unschedule + schedule.

## Live verification (real account)

- Scheduled a workout, then `get_scheduled_workouts` returned a matching
  `scheduled_workout_id` equal to the calendar entry id.
- `unschedule_workout` removed the entry; the calendar was empty afterward and
  the workout template still appeared in `get_workouts`.
- Note: the GraphQL summaries index is eventually consistent — right after a
  write it can briefly lag; the normal "list then remove" flow is unaffected,
  and this is documented in the tool docstring.

## Tests

- Added 5 integration tests: `unschedule_workout` success/error,
  `unschedule_workouts` batch/partial-failure, and `get_scheduled_workouts`
  exposing `scheduled_workout_id`.
- 252 integration tests pass. One pre-existing failure
  (`test_upload_workout_accepts_secondary_target_type_with_null_primary`) is
  unrelated to this PR — it fails identically on `main` with these changes
  stashed.

### DIFF
--- a/src/garmin_mcp/workouts.py
+++ b/src/garmin_mcp/workouts.py
@@ -410,6 +410,9 @@ def _curate_scheduled_workout(scheduled: dict) -> dict:
 
     summary = {
         "date": scheduled.get('scheduleDate'),
+        # Calendar-entry id (distinct from workout_id). Pass this to
+        # unschedule_workout to remove the entry from the calendar.
+        "scheduled_workout_id": scheduled.get('scheduledWorkoutId'),
         "workout_uuid": scheduled.get('workoutUuid'),
         "workout_id": scheduled.get('workoutId'),
         "name": scheduled.get('workoutName'),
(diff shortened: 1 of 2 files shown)


valdugay:
Hi @Taxuspt 👋

I originally built this for my own training setup — I wanted to remove a workout from my Garmin calendar (not delete the template itself), and then saw #132 asking for exactly that, so I figured I'd contribute it back rather than keep it local.

It adds `unschedule_workout` / `unschedule_workouts` and exposes the `scheduled_workout_id` that they need. I kept the scope focused and noted a couple of follow-ups in the description (a `(workout_id, calendar_date)` convenience wrapper and a `replace_scheduled_workout` tool) that I'm happy to do in separate PRs.

More than happy to get feedback and adjust anything — naming, the signature, scope — whatever fits the project best. Thanks for maintaining this! 🙏

PR #168, opened 2026-06-24
fix(workouts): validate calendar_date up front in schedule_workout/schedule_workouts

victory-c:
## What

`_validate_date` is already used across the workouts **read** tools (`get_scheduled_workouts`, `get_training_plan_workouts`) and inside the `_is_already_scheduled` idempotency pre-check — but the two **write** tools, `schedule_workout` and `schedule_workouts`, never validated their `calendar_date`.

Because `_is_already_scheduled` deliberately swallows its own errors (`except Exception: return False`, so the idempotency check can't block a legitimate schedule), a malformed date falls straight through to the real `POST workout-service/schedule/{id}` with a garbage body. The call is doomed, and the caller gets back an opaque `HTTP 4xx` / Garmin error string instead of a clear "wrong format" message.

## Change

Validate `calendar_date` up front in both tools, before any network call, matching each tool's existing failure style:

- **`schedule_workout`** — returns a structured `{"status": "failed", …, "message": "Invalid calendar_date '…': expected YYYY-MM-DD"}` and never POSTs.
- **`schedule_workouts`** — per item, rejects a bad date right next to the existing `calendar_date is None` guard (same `status: "failed"` shape, then `continue`). This also short-circuits **before** the inline `workout_data` upload, so a malformed date no longer triggers a wasted upload.

No happy-path change: valid `YYYY-MM-DD` dates flow exactly as before.

## Tests (+3)

- `test_schedule_workouts_rejects_invalid_date` — bad date → `failed`, `post` not called.
- `test_schedule_workouts_invalid_date_skips_inline_upload` — bad date on an inline item → no `upload_workout`, no `post`.
- `test_schedule_workout_rejects_invalid_date` — covers the single-workout tool (which previously had no tests) → `failed`, and neither `post` nor the GraphQL pre-check is called.

## Validation

```
uv sync
uv run pytest tests/integration tests/unit   # 388 passed (was 385)
```

### DIFF
--- a/src/garmin_mcp/workouts.py
+++ b/src/garmin_mcp/workouts.py
@@ -961,6 +961,16 @@ def register_tools(app):
             calendar_date: Date to schedule the workout in YYYY-MM-DD format
         """
         try:
+            _validate_date(calendar_date, "calendar_date")
+        except ValueError as e:
+            return json.dumps({
+                "status": "failed",
+                "workout_id": workout_id,
+                "scheduled_date": calendar_date,
+                "message": str(e),
+            }, indent=2)
+
+        try:
             if _is_already_scheduled(workout_id, calendar_date):
                 return json.dumps({
                     "status": "success",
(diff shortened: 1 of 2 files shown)


Taxuspt:

PR #175, opened 2026-06-26
fix(training): repair get_training_load_trend for device-keyed status data

kaosensei:
### Problem

`get_training_load_trend` returns `vo2_max` only — CTL, ATL, ACWR, training_status are all missing.

Root cause: `mostRecentTrainingStatus.latestTrainingStatusData` is a **dict keyed by device ID**, not the DTO directly:

```python
# actual structure
data["mostRecentTrainingStatus"]["latestTrainingStatusData"]
# == {"3620729386": {...device-specific status data...}}
```

The existing code does `.get("latestTrainingStatusData", {}).get("acuteTrainingLoadDTO", {})` which yields an empty dict every time, so every load field returns `None`.

Additionally, the device data no longer wraps training status fields in a `trainingStatusDTO` sub-object — `trainingStatus`, `trainingStatusFeedbackPhrase`, and `fitnessTrend` are now flat on the device record.

### Fix

- Iterate the device map and prefer the primary training device (fall back to the first entry), mirroring the pattern already used by `get_training_status`.
- Read the flattened status keys directly (no more `trainingStatusDTO` lookup).
- Use `or {}` throughout to coalesce explicit `None` (Garmin returns `null` for sections the user has no data in, which breaks chained `.get(key, {})`).

Also exposes `acwr_percent` and the chronic-load tunnel (`optimal_chronic_load_min`/`max`), which are useful for plotting the PMC and were already extracted by `get_training_status`.

### Related

The same device-keyed-dict pattern was flagged out-of-scope in #173 — this PR addresses it independently. The two PRs touch different functions and can be merged in any order.

### Verification

Local e2e on 2026-06-23 → 2026-06-25 now returns full `atl / ctl / tsb / acwr / acwr_status / acwr_percent / optimal_chronic_load_min / optimal_chronic_load_max / training_status / training_status_code / fitness_trend / vo2_max` per day. Unit + integration tests 385/385 pass.

### DIFF
--- a/src/garmin_mcp/training.py
+++ b/src/garmin_mcp/training.py
@@ -734,12 +734,25 @@ def register_tools(app):
             try:
                 data = garmin_client.get_training_status(date_str)
                 if data:
-                    status_data = (
-                        data.get("mostRecentTrainingStatus", {})
-                        .get("latestTrainingStatusData", {})
-                    )
-                    atl_dto = status_data.get("acuteTrainingLoadDTO", {})
-                    vo2_data = data.get("mostRecentVO2Max", {}).get("generic", {})
+                    # latestTrainingStatusData is a dict keyed by device ID, not
+                    # the DTO directly. Pick the primary training device when
+                    # available, fall back to the first device. Use `or {}` to
+                    # coalesce explicit nulls (Garmin returns None for sections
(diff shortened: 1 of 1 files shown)


Taxuspt:

PR #177, opened 2026-06-26
fix: handle null API values in get_training_status

### LINKED ISSUE #172: get_training_status crashes with NoneType when heatAltitudeAcclimationDTO is null

phdelodder:
## Problem
Several fields returned by Garmin's training status API can be
explicitly `null` — for example:
- `mostRecentVO2Max` for users without VO2max tracking data
- `mostRecentTrainingLoadBalance` for users without heat/altitude
  training history
Using `x.get(key, {})` does **not** guard against this. The default
`{}` only triggers when the key is **absent**. When the key is present
with a `null` value, `.get(key, {})` returns `None`, and the
subsequent `.get(...)` call raises: 

`'NoneType' object has no attribute 'get'`

This makes `get_training_status` (and tools that call it internally:
`get_training_load_trend`, `get_vo2max_trend`) completely unusable for
affected users.

Closes #172

## Fix

Replace all affected chained dict lookups with `(x.get(key) or {})`.
The `or {}` handles both cases — key absent and key present with
`null` — so the subsequent `.get(...)` always receives a dict.

## Affected locations

8 occurrences across 3 functions in `training.py`:
- `get_training_status` — 6 occurrences
- `get_training_load_trend` — 2 occurrences
- `get_vo2max_trend` (internal loop) — 1 occurrence

## Testing

Verified on a Garmin account where `mostRecentVO2Max` was previously
returned as `null` — all three tools now return data correctly.

### DIFF
--- a/src/garmin_mcp/training.py
+++ b/src/garmin_mcp/training.py
@@ -493,8 +493,11 @@ def register_tools(app):
                 return f"No training status data found for {date}."
 
             # Extract from nested structure
-            recent_status = status.get("mostRecentTrainingStatus", {})
-            latest_data = recent_status.get("latestTrainingStatusData", {})
+            # Use `(x.get(key) or {})` instead of `x.get(key, {})` so that
+            # explicit null values in the API response are treated as missing
+            # rather than causing `NoneType has no attribute 'get'` errors.
+            recent_status = (status.get("mostRecentTrainingStatus") or {})
+            latest_data = (recent_status.get("latestTrainingStatusData") or {})
 
             # Get first device data (usually the primary device)
             device_data = {}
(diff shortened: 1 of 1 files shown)


Taxuspt:

PR #180, opened 2026-07-01
Fix Garmin Connect China authentication

mtunique:
# Fix Garmin Connect China authentication

## Summary

Fix Garmin Connect China pre-authentication by applying a small compatibility
patch around `garminconnect==0.3.2`.

The upstream client is already domain-aware for many Garmin China URLs, but the
auth flow still has two China-specific issues:

- login tries the mobile flow first, whose service URL is not domain-aware
- DI OAuth token exchange uses the global `diauth.garmin.com` endpoint instead
  of `diauth.garmin.cn`

This patch applies the compatibility fix before the MCP server, auth CLI, or
token validation instantiate `Garmin`.

## Changes

- Add `garmin_cn_auth_patch.apply_garmin_cn_auth_patch()`
- Guard the compatibility patch so it only applies to `garminconnect==0.3.2`
- For `garmin.cn`, prefer portal login, then widget login, before mobile login
- For `garmin.cn`, rewrite DI OAuth POSTs to
  `https://diauth.garmin.cn/di-oauth2-service/oauth/token` without mutating the
  upstream module's global `DI_TOKEN_URL`
- Apply the patch in the server entrypoint, auth CLI, and token utilities
- Add unit tests covering the China login strategy order and DI endpoint

## Validation

- `python -m pytest tests/unit/test_garmin_cn_auth_patch.py`
- `python -m pytest tests/unit/test_auth_cli.py tests/unit/test_token_utils.py`
- `python -m pytest tests/unit` (`86 passed`)
- Live read-only verification against Garmin Connect China using saved OAuth
  tokens:
  - token login succeeded
  - `get_full_name()` returned the expected profile
  - `get_stats("2026-06-30")` returned daily metrics

### DIFF
--- a/src/garmin_mcp/__init__.py
+++ b/src/garmin_mcp/__init__.py
@@ -11,6 +11,10 @@ from mcp.server.fastmcp import FastMCP
 
 from garminconnect import Garmin, GarminConnectAuthenticationError, GarminConnectConnectionError, GarminConnectTooManyRequestsError
 
+from garmin_mcp.garmin_cn_auth_patch import apply_garmin_cn_auth_patch
+
+apply_garmin_cn_auth_patch()
+
 # Import all modules
 from garmin_mcp import activity_management
 from garmin_mcp import health_wellness
--- a/src/garmin_mcp/auth_cli.py
+++ b/src/garmin_mcp/auth_cli.py
@@ -13,6 +13,7 @@ import base64
 import requests
 from garminconnect import Garmin, GarminConnectAuthenticationError, GarminConnectConnectionError, GarminConnectTooManyRequestsError
 
+from garmin_mcp.garmin_cn_auth_patch import apply_garmin_cn_auth_patch
 from garmin_mcp.token_utils import (
     get_token_path,
     get_token_base64_path,
@@ -20,6 +21,8 @@ from garmin_mcp.token_utils import (
     validate_tokens,
     get_token_info,
 )
+
+apply_garmin_cn_auth_patch()
 
 
 def _secure_token_dir(path: str) -> None:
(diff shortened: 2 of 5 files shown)


pn:
Checked against current `garminconnect` and the existing upgrade in #214. This patch becomes redundant once #214 lands: that PR already bumps `garminconnect` to 0.3.7, raises the project floor to Python 3.12, regenerates `uv.lock`, and updates the CI matrix and documentation.

Upstream 0.3.4 made the DI token endpoint domain-aware, and 0.3.5 did the same for mobile service URLs. In 0.3.7 the old `.com` module constants have no call sites; authentication reads per-instance URLs derived from `domain`. `Client(domain="garmin.cn")` yields `diauth.garmin.cn`, `mobile.integration.garmin.cn/gcm/{ios,android}`, and `connect.garmin.cn/app`.

Compatibility: 0.3.2 → 0.3.7 removes no `Garmin` method or stored instance attribute, and the token-store fields are unchanged, so existing tokens remain readable. The current repository suites pass against 0.3.7 (126 unit + 354 mocked integration tests), although they mock the client and do not exercise a real authentication exchange.

Two remaining notes for #214:

- `dxt/manifest.json` still advertises Python `>=3.10` and should be aligned with the new `>=3.12` project requirement. The 3.12 floor began in `garminconnect` 0.3.3, so there is no CN-fixed release that retains Python 3.10/3.11.
- Upstream kept the mobile-first strategy order, and `GarminConnectAuthenticationError` aborts the chain. I could not test a live CN login, so if the CN mobile flow fails specifically that way there will be no portal fallback. The low-level `client.skip_strategies = {"mobile+cffi", "mobile+requests"}` provides that control without a monkey patch.

I recommend treating #214 as superseding this PR rather than merging the compatibility patch. The tests here also encode portal-before-mobile behavior and would fail after the upgrade because the patch intentionally version-gates itself to 0.3.2.

PR #182, opened 2026-07-06
Harden HTTP transport bind and token file permissions

hclpush:
## Summary

Two local-exposure hardening fixes for the HTTP transport and OAuth token storage. No behavior change for existing stdio users; all 81 unit tests pass.

## 1. HTTP transport no longer binds to all interfaces by default

`GARMIN_MCP_HOST` defaulted to `0.0.0.0`, and the HTTP transport performs no authentication of its own. Enabling an HTTP transport (e.g. the quickstart `GARMIN_MCP_TRANSPORT=streamable-http garmin-mcp`) therefore exposed **full read *and write* access to the user's Garmin account** — health data, GPS history, plus mutating tools (upload/delete workouts and courses, delete food logs) — to anything that could route to the host.

This changes the default to `127.0.0.1`. External binding is now an explicit opt-in: set `GARMIN_MCP_HOST=0.0.0.0` when the endpoint is fronted by an authenticating reverse proxy (as the README already recommends). Containers/k8s deployments that need external reachability set the env var explicitly, so the secure posture is the default and the exposed posture is deliberate.

## 2. OAuth token files hardened in the server auto-auth path

`auth_cli.py` already set `0700`/`0600` on written tokens, but the server's own fallback path in `init_api()` (`__init__.py`) dumped the token directory and wrote the base64 token file with **no `chmod`**. Under a typical `022` umask these land world-readable — and they are ~6-month bearer credentials to the whole account, so any other local user on a multi-user host could read and reuse them.

This adds a shared `secure_token_dir()` helper in `token_utils.py`, applies it right after `garmin.client.dump()` in the server path, and adds `chmod 0600` on the base64 file. `auth_cli.py`'s private helper now aliases the shared one so both auth paths harden identically.

## Changes
- `src/garmin_mcp/__init__.py` — default host `127.0.0.1`; secure tokens in `init_api()`
- `src/garmin_mcp/token_utils.py` — new shared `secure_token_dir()` helper
- `src/garmin_mcp/auth_cli.py` — reuse the shared helper (DRY)
- `README.md` + inline docs — document the new default
- `tests/unit/test_transport_config.py` — updated default-host assertion

## Testing
`uv run pytest tests/unit -q` → **81 passed**.

## Also noticed (not in this PR — happy to follow up)
- The base64 token file is a second full copy of the credentials on disk, but the code that reads it back is commented out (`__init__.py`), so it's an unused secondary secret store worth removing.
- `upload_course(gpx_path)` reads any local `.gpx`-suffixed file and POSTs it to Garmin; since MCP tool args are model-controlled this is a minor exfil surface worth a note.
- `.github/workflows/security.yml` is named "Security Checks" but runs no actual vulnerability scan (there's a `# Add pip-audit or safety scan here` placeholder). Wiring in `pip-audit` would make it do what its name implies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,7 +397,7 @@ File-based secrets are useful in certain environments, such as inside a Docker c
 By default the server communicates over **stdio**, which is what Claude Desktop, the MCP Inspector, and most local clients expect. To serve over **HTTP** instead (e.g. when running in a container or Kubernetes), set the transport via environment variables:
 
 - `GARMIN_MCP_TRANSPORT`: `stdio` (default), `streamable-http`, or `sse`
-- `GARMIN_MCP_HOST`: bind address for HTTP transports (default `0.0.0.0`)
+- `GARMIN_MCP_HOST`: bind address for HTTP transports (default `127.0.0.1`; set to `0.0.0.0` only when the endpoint is fronted by an authenticating reverse proxy)
 - `GARMIN_MCP_PORT`: bind port for HTTP transports (default `8000`)
 
 ```bash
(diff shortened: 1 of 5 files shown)


Taxuspt:

PR #186, opened 2026-07-13
Fix explicit cycling watt targets using power.lap

gonced8:
## Summary

- use Garmin target type ID `9` / `power.lap` for explicit cycling watt targets
- keep ID `2` / `power.zone` for cycling zones and ID `6` / `pace.zone` for pace targets
- accept finite, non-negative watt ranges, including recovery ranges beginning at `0 W`, while rejecting booleans, missing/reversed bounds, negative values, and non-finite values
- document the canonical mapping across `upload_workout`, `upload_workouts`, `schedule_workouts`, the README, and DXT metadata
- add `workout://templates/cycling-power-intervals` and rebuild `garmin-mcp.dxt`

## Why

Garmin treats the numeric target type ID as authoritative. Explicit cycling watts therefore need ID `9` / `power.lap`; using ID `6` is interpreted as `pace.zone` and loses the intended power target.

Current upstream `python-garminconnect` defines [`TargetType.POWER_LAP = 9`](https://github.com/cyberjunky/python-garminconnect/blob/2ae0eb5d6e56a13bc0194e229992c3a7d855253c/garminconnect/workout.py#L75-L88). This project remains pinned to `garminconnect==0.3.2`, whose [`TargetType` does not yet include `POWER_LAP`](https://github.com/cyberjunky/python-garminconnect/blob/0.3.2/garminconnect/workout.py#L67-L78); this PR intentionally does not change dependencies.

A Garmin upload/fetch round-trip also confirmed that ID `9` / `power.lap` preserves the explicit watt range.

## Impact

Clients can safely upload cycling workouts with explicit power targets, including zero-watt recovery lower bounds. The new resource provides a complete warmup, repeated work/recovery, and cooldown example using the canonical repeat-group condition.

## Testing

- `python -m pytest -q tests/test_workout_power_targets.py tests/test_workout_templates.py tests/integration/test_workouts_tools.py` — 92 passed
- `python -m pytest -q -m "not e2e"` — 416 passed, 19 deselected
- verified `dxt/manifest.json` matches the manifest embedded in `garmin-mcp.dxt` byte-for-byte


gonced8:
Closing this in favor of #194. Its live Garmin Connect upload/fetch/FIT/device validation confirms that explicit cycling watt ranges should use ID 2 / power.zone with targetValueOne and targetValueTwo, producing 3-second power guidance. ID 9 / power.lap is valid Garmin behavior, but represents lap-average power rather than the intended current-power target. Thanks to pn for the stronger end-to-end validation.

PR #191, opened 2026-07-18
Expose strokeType/equipmentType/drillType in curated swim workout steps

chiouhao:
_curate_workout_step in workouts.py extracts targetType/secondaryTargetType generically for all sports, but never reads strokeType, equipmentType, or drillType -- so get_workout_by_id silently drops stroke and equipment data for every swim workout. Confirmed via a raw connectapi pull against workout-service/fbt-adaptive/{uuid} that Garmin's backend returns these as clean, reliable nested dicts (e.g. {"strokeTypeKey": "free"}, {"equipmentTypeKey": "pull_buoy"}), fully independent of the ambiguous secondaryTargetValueOne field. This patch surfaces them the same way category/exerciseName are already surfaced for strength workouts. No behavior change for non-swim sports (fields are simply absent/null there).

### DIFF
--- a/src/garmin_mcp/workouts.py
+++ b/src/garmin_mcp/workouts.py
@@ -318,7 +318,19 @@ def _curate_workout_step(step: dict) -> dict:
         zone_field='secondaryZoneNumber',
         prefix='secondary_',
     )
-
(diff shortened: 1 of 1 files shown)


Taxuspt:

PR #192, opened 2026-07-20
Update Dockerfile

gmac7oaks:


Taxuspt:
Thanks for the PR! The  file referenced in the Dockerfile change isn't included in the PR, so the image would fail to build as-is. Could you add the  to the PR, or describe what it should contain? Once it's included, happy to merge this.

Taxuspt:
This PR references `entrypoint.sh` in the `COPY` and `ENTRYPOINT` instructions but the file itself is not included in the diff. Without `entrypoint.sh` the Docker image will fail to build. Could you add the missing file to this PR?

Taxuspt:
This Dockerfile change points ENTRYPOINT at /app/entrypoint.sh, but no entrypoint.sh is added anywhere in this PR (or elsewhere in the repo), so the image would fail to build/start. Closing — happy to take a look if you push the missing script.

Issue #136, opened 2026-06-11
Publish versioned Docker images (GHCR or Docker Hub)


skyf0l:
First off, thanks for maintaining this, it's a really useful project! 🙏

There's a `Dockerfile` but no published image, so everyone has to `docker build` locally. Could you publish versioned images on release?

  - **GHCR (`ghcr.io/taxuspt/garmin_mcp`)**: no extra secrets (uses the built-in `GITHUB_TOKEN`), lives next to the repo, multi-arch is easy. You'd just flip the package to public once. Slightly less "default" than Docker Hub for some users.
  - **Docker Hub**: most familiar (`docker pull taxuspt/garmin-mcp`), best discoverability. Needs you to add `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` secrets, and anonymous pulls hit Docker Hub's rate limits.

  My suggestion: **GHCR** (zero secrets, easiest for you) - or both if you want the Docker Hub visibility.

PR #170, opened 2026-06-24
ci: publish multi-arch Docker images to GHCR on release (closes #136)

victory-c:
## What

Closes #136. There's a `Dockerfile` but no published image, so everyone has to `docker build` locally. This adds a workflow that builds and pushes a versioned, multi-arch image to the GitHub Container Registry on each release.

Per the issue, this uses **GHCR** — zero extra secrets (the built-in `GITHUB_TOKEN`), lives next to the repo, and multi-arch is easy.

## How it works

- **Trigger:** `release: [published]` (plus `workflow_dispatch` for manual runs).
- **Image:** `ghcr.io/taxuspt/garmin_mcp` (`docker/metadata-action` lowercases the repo owner).
- **Tags:** semver `{{version}}` and `{{major}}.{{minor}}`, plus `latest`.
- **Platforms:** `linux/amd64` and `linux/arm64` (via QEMU + Buildx), with GitHub Actions layer caching.
- **Auth:** `GITHUB_TOKEN` with `permissions: packages: write`.

Also documents pulling the prebuilt image in the README's Docker section.

## One-time maintainer step

GHCR packages created by Actions inherit the repo's visibility. After the first release publishes the image, flip the package to **public** once under *Packages → Package settings* if you want anonymous `docker pull`.

## Validation

- Workflow YAML parsed and structurally checked (release trigger, `packages: write`, QEMU/Buildx/login/metadata/build-push steps present).
- **Note:** the actual multi-arch build/push only runs in the upstream repo on a real release (a Docker daemon wasn't available locally to dry-run the build). The `Dockerfile` is the existing, unmodified one already used by `docker-compose`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #195, opened 2026-07-20
Add secure GHCR container publishing

gonced8:
## What changed

- build the existing Dockerfile for `linux/amd64` and `linux/arm64` on pull requests
- publish `latest` and `sha-<short-sha>` images to GHCR after pushes to `main`
- add OCI source and revision labels plus GitHub Actions caching

## Security

- keep pull-request builds in a read-only workflow with no package permission
- isolate `packages: write` in the `main`-push publishing workflow
- pin every referenced Action to a full commit SHA
- use the repository-scoped automatic `GITHUB_TOKEN`; no custom secrets are required

## Validation

- parsed both workflow files as YAML
- verified all Action references use full SHAs
- verified PR builds cannot publish and have no `packages: write` permission
- verified the publish workflow targets both architectures and only runs on `main` pushes

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,0 +1,31 @@
+name: Container build
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build multi-architecture image
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
+
+      - name: Set up Docker Buildx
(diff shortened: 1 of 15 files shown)


mknightz:
Review recommendation: **split this PR before merging.** It currently bundles two unrelated changes, and the bundle is what makes it hard to land:

1. **CI/GHCR publishing** (`container-build.yml`, `container-publish.yml`, Dockerfile hardening, README Docker docs) — good content, but it overlaps #170, which already implements release-tagged multi-arch GHCR publishing and closes #136. Two publish workflows would race on tags. The valuable unique piece here is the **SHA-pinned action references + PR-builds-without-publish-permissions split** — that hardening belongs folded into #170's workflow, not as a parallel pipeline.

2. **A 616-line `exercise_catalog.py` + workout-builder changes + `__init__` changes** — this is a strength-exercise-catalog *feature*, overlapping the open #251 (catalog tools + FIT catalog resolution), #238 (draft, catalog resolution), and #208 (structured strength creation). Merging it here would three-way-conflict with whichever catalog approach the maintainer picks.

Concrete path: extract the container workflow + hardening into #170 (or a small standalone CI PR), and let the catalog half be reviewed on its own against the other three catalog PRs. As bundled, the PR is effectively blocked on two unrelated decisions.

PR #197, opened 2026-07-21
fix: use reps for strength workout end condition instead of hardcoded time

### LINKED ISSUE #196: create_strength_workout never sets exerciseCategory and always uses a time-based end condition, ignoring reps
Built a strength workout via create_strength_workout (5 exercises, real names like "Goblet Squat," "Romanian Deadlift," with explicit reps values), then inspected it two ways:

Via get_workout_by_id — every step came back as a generic interval with a time-based end condition (not reps), despite reps being passed in.
Via the Garmin Connect web UI's step editor — each step's "Exercise → Type" field showed "Choose an Exercise" unresolved, with a validation warning "An exercise is required."
Root cause (in workout_builders.py, build_strength_json):

Each step only sets exerciseName (free text) — there's no exerciseCategory field anywhere in the payload. Garmin's real exercise picker needs both a category and a name from its internal catalog to resolve to an actual exercise (with animation/icon); a bare exerciseName string doesn't get matched to anything.
endCondition is hardcoded to "time" with endConditionValue = sets * 45 (a flat 45s/set guess) — the reps argument passed into the tool is only used in the description text, never as the actual end condition, even though Garmin's schema supports a reps-typed end condition (confirmed via the web UI's "Target Type" dropdown, which lists Reps as an option).
Impact: workouts created via this tool show up on the calendar and have the right structure/timing, but exercises don't render with real names/animations on the watch, and steps run as fixed timers rather than waiting for the prescribed rep count.

Possible fix: populate exerciseCategory from a real category/name mapping, and use conditionTypeKey: "reps" when a reps value is provided. The catalog itself isn't publicly documented as far as I can tell — would need reverse-engineering via a network capture while manually building a workout with real exercises selected in Garmin Connect.

victory-c:
This PR addresses Issue #196 where `create_strength_workout` ignored the user-provided `reps` count and defaulted to a hardcoded time-based end condition.

### Changes:
- Changed `conditionTypeId` from `2` (time) to `10` (reps).
- Assigned the actual `reps` value to `endConditionValue`.
- Added a default `category: "UNASSIGNED"` to the payload to satisfy Garmin Connect strict validation for strength exercises.

Closes #196.

### DIFF
--- a/src/garmin_mcp/workout_builders.py
+++ b/src/garmin_mcp/workout_builders.py
@@ -252,9 +252,10 @@ def build_strength_json(
             "stepOrder": step_order,
             "stepType": {"stepTypeId": 3, "stepTypeKey": "interval"},
             "description": f"{ex_name}: {sets} sets x {reps} reps",
-            "endCondition": {"conditionTypeId": 2, "conditionTypeKey": "time"},
-            "endConditionValue": float(sets * 45),  # rough estimate: 45s per set
+            "endCondition": {"conditionTypeId": 10, "conditionTypeKey": "reps"},
+            "endConditionValue": float(reps),
             "targetType": {"workoutTargetTypeId": 1, "workoutTargetTypeKey": "no.target"},
+            "category": "UNASSIGNED",
             "exerciseName": ex_name,
         })
         step_order += 1


Taxuspt:

PR #202, opened 2026-07-21
feat: support custom bpm heart-rate range in create_run_workout

mmccomiskey:
## Summary
- `create_run_workout` only accepted a named Garmin HR zone (`Z1`-`Z5`), but not every real training target lines up with a whole zone — e.g. a 136-148 bpm Zone 2 goal straddles Z2 (118-137) and Z3 (138-157). The watch would show "in range" for the entire zone (up to 157 bpm) even when the intended ceiling was lower, silently misleading in-workout HR feedback.
- Adds optional `hr_min`/`hr_max` params. When both are given, the workout targets an exact custom bpm range (`targetValueOne`/`targetValueTwo`) instead of a `zoneNumber` — a target shape Garmin Connect already supports natively (same `heart.rate.zone` target type), just not previously exposed by this tool.
- `hr_zone` remains the default, unchanged behavior when no range is given.

## Test plan
- [x] Existing unit/integration tests for `create_run_workout` / `build_run_json` pass unchanged (zone-based path untouched)
- [x] New unit tests: custom range builds `targetValueOne`/`targetValueTwo` with no `zoneNumber`; rejects a lone `hr_min`/`hr_max`; rejects `hr_min >= hr_max`
- [x] New integration test: tool call with `hr_min`/`hr_max` produces the expected uploaded JSON
- [x] Verified end-to-end against the live Garmin Connect API (not just mocks): uploaded a throwaway workout with a 136-148 bpm range, fetched it back, confirmed `targetValueOne=136.0`, `targetValueTwo=148.0`, `zoneNumber=None`, then deleted the test workout

### DIFF
--- a/src/garmin_mcp/workout_builders.py
+++ b/src/garmin_mcp/workout_builders.py
@@ -45,22 +45,54 @@ def _zone_number(zone: str) -> int:
     raise ValueError(f"Invalid hr_zone '{zone}'. Use Z1-Z5 or 1-5.")
 
 
+def _hr_target(
+    hr_zone: str,
+    hr_min: Optional[int],
+    hr_max: Optional[int],
+) -> tuple:
+    """Resolve HR target fields and a short description suffix.
+
+    Returns (target_extra_fields, description_suffix). If hr_min/hr_max are
+    both given, builds a custom bpm-range target (targetValueOne/targetValueTwo)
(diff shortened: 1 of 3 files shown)


Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #224. Thank you for the contribution @mmccomiskey!

PR #207, opened 2026-07-22
fix(weather): correct temperature unit + label wind for activity weather

tamcore:
The activity-weather endpoint's values are NOT uniformly in the account's display unit (contrary to the assumption in #163). Verified against the raw API and a local weather station:

- TEMPERATURE is returned in Fahrenheit with no unit field, regardless of account. A metric account saw raw Fahrenheit mislabeled 'C' (e.g. 57 tagged C, which is impossible; 57F = 13.9C). Fix: treat as Fahrenheit and convert to the account unit (metric -> C, statute_us -> F).
- WIND is already in the account unit (station cross-check: 8 = km/h for a metric account, not mph). So it is not converted; a wind_speed_unit label (km/h metric, mph statute_us) is added instead.

Applies to temperature, apparent_temperature, dew_point.

### DIFF
--- a/src/garmin_mcp/activity_management.py
+++ b/src/garmin_mcp/activity_management.py
@@ -655,14 +655,15 @@ def register_tools(app):
     async def get_activity_weather(activity_id: Union[int, str]) -> str:
         """Get weather data for an activity.
 
-        The temperature and dew_point values are in the unit matching the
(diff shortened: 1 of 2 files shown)


Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #223. Thank you for the contribution @gitgrave!

PR #208, opened 2026-07-24
feat: add structured strength activity creation and set updates

pn:
## Summary

Add two complementary MCP tools for structured sets on completed Garmin strength activities:

- `set_activity_strength_exercise_sets` safely replaces and verifies the complete set list of an existing `strength_training` activity
- `create_strength_training_activity` creates a completed private strength activity, attaches structured sets, and verifies the saved result

Both tools support active and rest sets, repetitions, external weight, duration, repeated sets, rest spacing, explicit offsets, and absolute start times.

## Dependency on #214

This PR is intentionally stacked on #214 because it relies on `garminconnect==0.3.7` for:

- the native `Garmin.set_activity_exercise_sets()` method rather than a private HTTP call
- the bundled English exercise catalog also exposed by `get_exercise_types`
- the upstream Python 3.12+ runtime requirement

#214 currently has two commits on the current `main` (`a16f057`), ending at
`210a5ab`. This branch is rebased onto that exact head before the #208 commits.
GitHub cannot use #214's branch directly as this PR's base because that branch
belongs to another contributor's fork.

Review only the #208 portion of the stack here:

https://github.com/pn/garmin_mcp/compare/210a5ab...b7b5d02

After #214 merges, this branch should be rebased onto the resulting `main` to
remove the now-duplicated dependency commits from the displayed diff.

## Why this is a separate capability

The related tools serve different Garmin objects:

- `get_activity_exercise_sets` reads sets from a completed activity
- `create_strength_workout` creates a planned workout that can be sent to a device
- `create_manual_activity` creates a basic completed activity without structured sets
- `get_exercise_types` exposes identifiers from Garmin's bundled catalog

This PR adds the missing safe write path for completed activity sets and composes it with manual activity creation.

## Updating an existing activity

`set_activity_strength_exercise_sets`:

- treats the operation as a full replacement rather than a partial patch
- verifies that the target exists and is explicitly a `strength_training` activity
- supports `dry_run=true` for validation and payload preview
- requires `confirm=true` before writing
- saves the previous set list before mutation
- delegates the write to `garminconnect.set_activity_exercise_sets`
- reads the result back and compares type, timing, repetitions, weight, and exercise identifiers
- retries one successful but stale/mismatched read after a short delay
- leaves read errors to `garminconnect`'s own retry/error policy instead of
  immediately repeating 4xx or rate-limited requests
- restores the previous set list by default when the write or verification still fails

## Creating a completed strength activity

`create_strength_training_activity`:

1. Validates activity metadata and the complete set timeline.
2. Returns a mutation-free preview when `dry_run=true`.
3. Requires `confirm=true` for creation.
4. Creates a private manual `strength_training` activity.
5. Reads only the documented top-level `activityId` from the create response.
6. Attaches the complete structured set list through the native library setter.
7. Reads the sets back and verifies all relevant fields, with one delayed retry
   when the first successful read is stale or mismatched.
8. Deletes the newly created activity if attachment or verification fails, unless `rollback_on_failure=false` was explicitly requested.

The result distinguishes creation, set-write, verification, and rollback
outcomes so callers can determine whether an incomplete activity may remain.
If the create request raises before returning an ID, the result explicitly
warns that Garmin may still have created an activity that requires manual
cleanup.

## Exercise catalog and naming

The implementation uses `garminconnect.exercises.EXERCISES`, the same bundled English catalog exposed by #214's `get_exercise_types`. It does not download or maintain a separate catalog.

- The `exercise` field matches English display names only, including conservative fuzzy matching.
- Technical identifiers require an exact `category` / `exercise_name` pair because bare identifiers can occur in multiple specialty categories.
- Common display names such as `squat`, `deadlift`, `push up`, and `pull up` resolve to their plain catalog entries.
- Incomplete specialty names such as `overhead press` and `ab twist` are rejected rather than silently mapped to sandbag or banded variants.
- Ambiguous or low-confidence matches are rejected rather than guessed.
- No handwritten multilingual alias table is maintained.
- `name` remains an accepted alias for `exercise_name` because Garmin's completed-activity payload uses `exercises[].name`.
- Supplying conflicting `name` and `exercise_name` values is rejected.
- Category-only catalog entries repeat the category as their workout exercise
  identifier. The completed-activity endpoint rejects that repeated value, so
  this boundary translates it to the endpoint's documented `name=null`
  representation while preserving specific sub-category identifiers.

## Validation and safety

- validates finite, non-negative durations and weights
- requires integer repetition and repeat counts
- treats explicit JSON `null` like omission for optional timing and set-type fields
- validates chronological ordering, non-overlapping sets, and activity-duration bounds
- derives the activity time zone when available and accepts an explicit override
- performs elapsed-time arithmetic in UTC so repeated sets stay correctly spaced across DST transitions
- converts set timestamps to Garmin's UTC representation
- converts kilograms to Garmin's internal gram representation
- accepts Garmin's bodyweight and missing-repetition normalizations during read-back
- rejects writes if the bundled catalog cannot validate the requested exercise
- uses activity deletion only as bounded rollback for an activity created by
  the same tool call; it does not expose a general destructive deletion tool

## Provenance

This remains an independent implementation. No source code or exercise catalog from the third-party GPL MCP server is included or copied.

## Verification

- Python 3.12.13: `556 passed`
- Python 3.13.13: `556 passed`
- focused structured-strength tests: `94 passed`
- cross-PR workout-builder and workout-tool regressions: `99 passed`
- `uv lock --check`
- `git diff --check`
- real bundled-catalog probes for safe common-name resolution and rejection of incomplete specialty names
- live Garmin endpoint probe: `SHOULDER_PRESS`, `LATERAL_RAISE`, and
  `OLYMPIC_LIFT` self-names each returned `400 Invalid Sub-Category`; the same
  known parent with `name=null` was accepted and read back as null
- clean stack ancestry: current `main` (`a16f057`) → exact #214 head (`210a5ab`) → #208
- prior live Garmin Connect validation during PR development covered disposable activity creation, structured set attachment, read-back verification, replacement, rollback cleanup, and final deletion


pn:
**Review via GitHub Copilot CLI** (model: Claude Opus 4.8) 🤖

Reviewed over two rounds — checked out the branch, read the module + tests, and ran the suites locally.

All three substantive findings are fixed and tested: float ints (`10.0`), recoverable set replacement (backup + rollback), and the missing `tzdata` dep. Remaining suggestions were intentionally declined by the author with sound reasoning (correctness/determinism/safety over convenience) — I agree.

New strength tests pass (67), no regressions. LGTM 👍

pn:
**Review via Claude Code (model: Claude Fable 5)** 🤖

**Verdict: ready to merge at 23ccc9c.**

**Scope & verification**
- Reviewed both tools (`set_activity_strength_exercise_sets`, `create_strength_training_activity`), the validation/matching/read-back helpers, tests, and docs.
- Confirmed repo conventions (`garmin_client.client.put("connectapi", …, api=True)` matches nutrition.py) and `garminconnect` integration: method signatures match, and HTTP ≥ 400 raises, so the rollback paths are actually reachable.
- Ran the full suite at the PR head on Python 3.10 and 3.14: **479 passed on both**.

**Key finding (fixed in 23ccc9c)**
- The initial revision was broken on Python 3.10 (minimum supported version, in the CI matrix): `datetime.fromisoformat` there rejects Garmin's single-digit fractional seconds (`…T08:00:00.0`), and the module also re-parses its own emitted timestamp format — 24 of the new tests failed on 3.10 while the pre-existing 405 passed. The fix (`_fromisoformat_compatible`) is correct: verified `.0`, `.5Z`, `.25-05:00`, and space-separated timestamps on 3.10, with >6-digit fractions still rejected consistently with 3.11+.

**Also addressed**
- `time_zone` now resolves explicit param → activity `timeZoneUnitDTO.unitKey` → UTC, and the dry-run preview reports the resolved zone.
- Explicit `exerciseSets: null` read-back is treated as an empty list; a missing key still aborts the pre-save.
- Tests added for offset placement, timezone-aware set `start_time`, activity-zone derivation, and null read-back; fuzzy-match thresholds extracted to named constants; README example no longer mixes `rest_seconds` spacing with explicit REST sets.

**Intentionally not changed — agreed**
- No auto-roll of past-midnight clock times: could mask typos; the limitation is now documented in both docstrings and the README.
- Nested `sets` repeat count kept: one canonical, documented name for a new API is reasonable, and it matches natural "3 sets of 10" phrasing.
- No live-Garmin test for the null case: CI stays deterministic; the defensive handling plus mock coverage suffice.

pn:
Rebased and refactored after #214 at `9c3faf2`.

The earlier review comments on this PR describe the previous `23ccc9c` implementation and are now superseded in three important respects:

- the branch requires Python 3.12+ through `garminconnect==0.3.7`
- structured set writes now delegate to the library's native `set_activity_exercise_sets()` method instead of calling the private HTTP client
- exercise matching now uses the bundled English catalog shared with `get_exercise_types`, with `category` / `exercise_name` as the documented cross-tool identifiers and `name` retained as an activity-payload alias

The branch is a logical stack:

```text
current upstream/main
└── #214 dependency replay
    └── #208 implementation and refactor
```

GitHub cannot set #208's base directly to #214 because #214's head is in another contributor's fork. Review only the #208 portion here:

https://github.com/pn/garmin_mcp/compare/077b746...9c3faf2

Verification after the refactor:

- Python 3.12.13: 488 unit/integration tests passed
- Python 3.13.13: 488 unit/integration tests passed
- `uv lock --check`
- `git diff --check`

This PR should not merge before #214. Once #214 lands, I will rebase #208 onto updated `main` to remove the replayed dependency commit.

pn:
Addressed the review in `2f6dfc1` after reproducing the matching findings against the real 1,527-entry bundled catalog.

Implemented:

- Free-form `exercise` input now matches English display names only. Bare technical identifiers no longer receive exact-match confidence.
- Real-catalog tests now cover `squat`, `deadlift`, `push up`, and `pull up`, plus rejection of incomplete specialty names such as `overhead press` and `ab twist`. The unit-wide synthetic catalog fixture was removed.
- Post-mutation verification reads get one immediate retry after an exception. A transient GET no longer reverts a successful write; persistent failure still follows the existing rollback guarantee.
- Repeat/offset arithmetic is performed in UTC, with a regression test across the Europe/Warsaw DST fallback.
- Explicit JSON `null` uses documented defaults for `set_type`, `duration_seconds`, and `rest_seconds`.
- Newly created activity IDs are accepted only from documented top-level response fields, preventing a nested unrelated ID from becoming a write/delete target.
- Missing activity type, zero summary duration, empty/invalid time zones, non-string activity names, and activity-ID response consistency were hardened.
- The per-set 1,527-entry dictionary allocation and the confusing generator variable shadowing were removed.

I intentionally did not apply a few suggestions literally:

- I did not add a `category == exercise_name` tie-break rule. Exact display-name matching resolves the common lifts correctly without relying on an identifier convention that Garmin could change.
- I did not keep unsafe bare identifiers under an `identifier_only` match type. Rejecting them and requiring `category` / `exercise_name` is safer than asking callers to notice a warning after an otherwise confident resolution.
- I did not disable rollback when both verification reads fail. One retry handles the transient-network case while preserving the tool's documented transactional behavior when the final state remains unknown.
- I did not add another cached catalog index. The implementation now does a simple allocation-free linear lookup over ~1,527 entries; for an activity-sized set list this is negligible and avoids another cache with invalidation/coupling concerns.
- I kept `tzdata` unconditional because slim Linux images can lack a system zone database too; limiting it to Windows would make behavior depend on the container base image.
- A repository-wide `confirm` convention remains out of scope for this feature PR.

Verification:

- Python 3.12.13: 503 passed
- Python 3.13.13: 503 passed
- Focused structured-strength tests: 91 passed
- `uv lock --check`
- `git diff --check`

PR #209, opened 2026-07-24
feat: create completed strength activities with structured sets

pn:
## Dependency

This PR is intentionally stacked on #208, which adds the validated strength-set writer used here. Please review and merge #208 first. Until then, GitHub shows both commits in this PR; after #208 merges, this PR's diff against `main` will reduce to the activity-creation commit.

For the incremental diff, use the
[fork branch comparison](https://github.com/pn/garmin_mcp/compare/feat/set-strength-exercise-sets...feat/create-strength-training-activity).

## Summary

- add `create_strength_training_activity` for creating a completed private Garmin strength activity with structured exercise sets
- reuse the same validated set format, catalog-based exercise matching, timing rules, and read-back verification introduced in #208
- support a mutation-free dry run before creation
- roll back a newly created activity by default if attaching or verifying its sets fails

## Why this does not duplicate existing tools

`create_manual_activity`, added in #157, creates the basic completed activity record only. It does not attach structured strength sets or verify them.

`create_strength_workout` and the strength changes merged through #201 create or correct a planned workout for a watch. They do not create a completed activity in the user's activity history.

This tool composes the existing manual-activity API with the completed-activity set writer from #208 and adds transactional failure handling around the two-step private Garmin API flow.

## Creation flow

1. Validate the activity name, duration, IANA time zone, local start time, set timeline, exercise matches, repetitions, and weights.
2. Return a complete preview when `dry_run=true`.
3. Require `confirm=true` for a real mutation.
4. Create a private manual activity with type `strength_training`.
5. Extract the returned activity ID, including nested response forms.
6. Attach the full structured set list.
7. Read the sets back and compare all relevant fields.
8. Delete the new activity when set attachment or read-back verification fails, unless `rollback_on_failure=false` was explicitly requested.

The result distinguishes creation, set-write, verification, and rollback outcomes so callers can tell whether any incomplete activity remains.

## Testing

- `python -m pytest tests/integration tests/unit -q` — 459 passed for the stacked branch
- 54 targeted unit and FastMCP integration tests for the two structured strength tools
- `uv lock --check`
- `git diff --check`
- live Garmin Connect end-to-end validation:
  - dry-run preview
  - create a disposable private `strength_training` activity
  - attach four active/rest, bodyweight, and weighted sets
  - independently read back the activity type, name, timing, repetitions, and weight
  - replace the set list through #208 and verify the changed values
  - delete the activity and verify that a subsequent read returned 404

The live test activity was clearly marked as disposable and was removed successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ This MCP server implements **110+ tools** covering ~90% of the [python-garmincon
 - ✅ High-Level Workout Builders (4 tools) - create and schedule workouts without writing JSON
 - ✅ Courses (3 tools) - list / upload GPX as course / delete course
 - ✅ Activity Analysis (2 tools) - FIT file parsing, Power Duration Curve; requires power meter and/or Di2
-- ✅ Structured Strength Activities (1 tool) - safely replace and verify completed activity sets
+- ✅ Structured Strength Activities (2 tools) - create completed strength activities or safely replace and verify their sets
 - ✅ Activity File Downloads (2 tools) - download activity files in FIT, GPX, TCX, or CSV format
 
 > **Note:** Activity Analysis tools require a compatible power meter (e.g., Garmin Rally, Favero Assioma, PowerTap P1) and/or Shimano Di2 / SRAM eTap electronic shifting. The `fitparse` dependency is installed automatically.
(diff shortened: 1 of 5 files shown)


pn:
Superseded by #208, which now contains both commits so the shared implementation and the dependent activity-creation flow can be reviewed and merged atomically.

PR #211, opened 2026-07-25
fix(workouts): canonicalize running pace target bounds

### LINKED ISSUE #210: Running pace.zone targetValueOne/Two accepted on upload but show as 0:00 on device and in preview
**Bug:** Uploading a running workout step with a `pace.zone` target (`workoutTargetTypeId: 6`, `workoutTargetTypeKey: "pace.zone"`, `targetValueOne`/`targetValueTwo` in m/s) succeeds with no error, and `get_workout_by_id` even echoes `target_type: pace.zone` back -- but the actual speed bounds never make it to the athlete. Confirmed on a real device, not just the Connect app preview.

**Repro payload** (used on both a distance-ended 400m interval step and a time-ended 25-min interval step, same result both times):
```json
{
  "type": "ExecutableStepDTO",
  "stepOrder": 3,
  "stepType": {"stepTypeId": 3, "stepTypeKey": "interval"},
  "endCondition": {"conditionTypeId": 3, "conditionTypeKey": "distance"},
  "endConditionValue": 400,
  "targetType": {"workoutTargetTypeId": 6, "workoutTargetTypeKey": "pace.zone"},
  "targetValueOne": 4.651,
  "targetValueTwo": 4.878
}
```

**Observed in Garmin Connect mobile app (screenshot of the step detail):**
> Laufen - 0,4 km
> Pace - 0:00 min/km (0:00 - 0:00 min/km)
> Zielpace 82-86 Sek.  (this last line is just my own free-text description field, not derived from the target)

So the app clearly resolves the target type correctly (it renders a "Pace" row at all), but the low/high bounds both read 0:00.

**Confirmed on-device too:** athlete ran the workout on a real Garmin watch. During the interval step, the device showed only the remaining time/distance -- no pace target zone was displayed or alerted on, even though the step type and description were correct.

**What I ruled out already (source-level, not just guessing):**
- `_fix_hr_zone_step` in `workouts.py` only fires for `heart.rate.zone`, not `pace.zone` -- not the cause.
- No unit conversion happens anywhere between the MCP tool and `garminconnect.upload_workout`, which just does a raw POST of the JSON as-is -- values aren't being mangled in transit.
- Checked `garminconnect/workout.py`'s typed `TargetType`/`ExecutableStep` models -- no pace-specific handling exists there either; `targetType` is just a passthrough dict.

**Not the same as prior closed issues:**
- #129 was about `workoutTargetTypeId: 6` + wrong key (`heart.rate`) getting silently resolved as `pace.zone` -- not applicable here, our key/ID pair (6/pace.zone) is already the canonical correct pairing per that issue's own findings.
- #89 was about swim workouts using `secondaryTargetType`/`secondaryTargetValueOne`/`Two` instead of the primary target fields -- also a different field-shape issue, and swim-specific.

**Question for whoever investigates:** is it possible running `pace.zone` targets actually need to go through `secondaryTargetType`/`secondaryTargetValueOne`/`secondaryTargetValueTwo` as well (the way swim workouts apparently do per #89), rather than the primary `targetType`/`targetValueOne`/`targetValueTwo` fields? That would be consistent with both bugs sharing a root cause even though the surface symptom differs.

**Workaround in use for now:** rely on the free-text `description` field (e.g. "Zielpace 82-86 Sek.") to convey the target to the athlete, since the actual pace.zone widget doesn't work.

pn:
> **Status after real-device verification:** the reported failure could not be reproduced. Both the reported and canonical bound orders displayed and enforced the correct pace range on a Forerunner 255. This PR is being closed because its normalization is not demonstrated to fix #210 and would add behavior without a confirmed need.

## Summary

- canonicalize primary running `pace.zone` ranges before upload so Garmin receives the faster (higher m/s) bound in `targetValueOne` and the slower bound in `targetValueTwo`
- apply the same behavior to `upload_workout`, `upload_workouts`, and inline `schedule_workouts` uploads, including steps nested in repeat groups
- leave canonical, incomplete, zone-based, non-numeric, swimming-secondary, and non-running targets unchanged
- document the running primary-target order and the separate swimming secondary-target field family

## Original hypothesis

Issue #210 reports that an ascending running speed range is accepted by Garmin, but Garmin Connect shows `0:00` bounds and the watch provides no pace guidance.

Garmin-authored running workouts consistently store pace ranges in pace order:

```text
targetValueOne = faster boundary (higher m/s)
targetValueTwo = slower boundary (lower m/s)
```

Live API probing showed that Garmin preserves the submitted order in stored workout JSON, while FIT export normalizes either order to equivalent low/high speed fields. This PR tested whether canonicalizing MCP output would avoid the reported presentation/device failure.

## Verification result

Two otherwise identical workouts were uploaded, scheduled, synced, and tested on a Forerunner 255:

- reported order: `4.651` / `4.878`
- canonical order: `4.878` / `4.651`

Garmin Connect Mobile and the watch displayed the correct approximately `3:25–3:35/km` range for both. Target guidance was active for both. The `0:00` preview and missing watch target from #210 did not occur.

No `garmin_mcp` change was merged between the issue report and this verification that could explain the different result. A Garmin service/app change remains possible but unverified. The remaining likely variables are the reporter's complete workout payload, device/firmware, app version/platform, or sync state.

## Implementation and tests retained in this branch

- normalization is restricted to primary `pace.zone` targets in running segments
- segment sport takes precedence, with workout sport used as fallback
- malformed `sportType` shapes retain their previous pass-through behavior
- validation runs before normalization, so rejected payloads are not mutated
- swimming pace targets in `secondaryTargetType` fields are not modified
- `416` unit and integration tests passed
- a live MCP → Garmin JSON → FIT test passed and cleaned up its temporary workout

These tests establish that the implementation is internally safe; they do not establish that normalization is necessary or fixes #210.

## Interaction with PR #194

This change is functionally independent from #194, which corrects cycling watt target encoding. Both branches touch target-type documentation, but #194 remains unrelated to the reported running-pace behavior.

Related to #210.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,24 @@ For a named Garmin HR zone, use the same target type with `zoneNumber` instead:
   "zoneNumber": 3
 }
 ```
+
+Running pace ranges use target type ID `6` with `pace.zone`. The values are
+speeds in metres per second, in Garmin's pace order: the faster (higher) speed
+goes in `targetValueOne`, and the slower (lower) speed in `targetValueTwo`.
+`upload_workout` and `upload_workouts` normalize reversed running bounds before
+upload.
+
+```json
+{
+  "targetType": {
+    "workoutTargetTypeId": 6,
+    "workoutTargetTypeKey": "pace.zone"
+  },
+  "targetValueOne": 4.878,
+  "targetValueTwo": 4.651
+}
+```
+
 ## One-click Install (Claude Desktop)
 
 The easiest way to add this server to Claude Desktop is via the `.dxt` Desktop Extension file — no JSON editing required.
--- a/src/garmin_mcp/workout_templates.py
+++ b/src/garmin_mcp/workout_templates.py
@@ -262,7 +262,7 @@ WORKOUT_STRUCTURE_REFERENCE = {
         "1": {"workoutTargetTypeKey": "no.target", "description": "No specific target"},
         "2": {"workoutTargetTypeKey": "power.zone", "description": "Cycling power zone 1-7 (use zoneNumber; based on FTP %). Do NOT use for absolute watt targets."},
         "4": {"workoutTargetTypeKey": "heart.rate.zone", "description": "Heart rate zone (use zoneNumber 1-5 for named zones, or targetValueOne/targetValueTwo for custom bpm range)"},
-        "6 (running/swim)": {"workoutTargetTypeKey": "pace.zone", "description": "Pace zone for running or swimming (use zoneNumber)"},
+        "6 (running/swim)": {"workoutTargetTypeKey": "pace.zone", "description": "Running pace range uses targetValueOne=fast speed and targetValueTwo=slow speed in m/s; swim pace is stored in secondary target fields"},
         "6 (cycling)": {"workoutTargetTypeKey": "power.between", "description": "Cycling absolute watt range (use targetValueOne=low_watts, targetValueTwo=high_watts). Use this instead of power.zone when targeting specific watts, NOT zone numbers."}
     },
     "sportType_values": {
--- a/src/garmin_mcp/workouts.py
+++ b/src/garmin_mcp/workouts.py
@@ -120,6 +120,53 @@ def _fix_repeat_group_step(step: dict) -> None:
         _fix_repeat_group_step(nested)
 
 
+def _normalize_running_pace_target_step(step: dict) -> None:
+    """Put running pace bounds in Garmin's native fast-to-slow order.
+
+    Garmin represents running pace ranges as speed in metres per second, but
+    unlike ordinary numeric ranges it expects the faster (larger) speed in
+    targetValueOne and the slower (smaller) speed in targetValueTwo.
+    Garmin-authored running workouts consistently use this canonical order.
+    """
+    target_type = step.get('targetType')
+    if (
+        isinstance(target_type, dict)
+        and target_type.get('workoutTargetTypeKey') == 'pace.zone'
+    ):
+        value_one = step.get('targetValueOne')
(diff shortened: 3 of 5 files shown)


pn:
Real-device follow-up for #210:

I uploaded and scheduled two otherwise identical workouts on a Forerunner 255:

- raw ascending bounds: `4.651` / `4.878`
- canonical faster-first bounds: `4.878` / `4.651`

Both displayed the correct range in Garmin Connect Mobile and on the watch, and both provided active target guidance. At `4:00/km`, the watch correctly showed the athlete as slower than the prescribed `3:25–3:35/km` range.

Therefore the original `0:00`/missing-target symptom was **not reproduced**, and reversed bound order does not appear causal on this account/device. This PR should be evaluated only as defensive canonicalization of MCP output, not as a demonstrated fix for #210.

pn:
Closing after the real-device A/B test changed the conclusion.

Both the reported ascending order (`4.651` / `4.878`) and the canonical faster-first order (`4.878` / `4.651`) displayed the correct range in Garmin Connect Mobile and on a Forerunner 255, with active target guidance in both cases. Therefore bound order is not a demonstrated cause of #210 on the tested account/device.

The implementation and tests are sound as defensive canonicalization, but that alone is not enough reason to add normalization across all upload paths. Merging it would introduce behavior without evidence that Garmin requires it or that it fixes the reported bug.

No `garmin_mcp` commit was merged between the July 24 report and the July 25 device test, so a repository change cannot explain why the issue did not reproduce. A Garmin-side change is possible but unverified.

I am preserving the branch and test results for reference, but closing the PR rather than presenting a speculative change as a fix.

PR #213, opened 2026-07-25
fix(workouts): omit invalid hardcoded category on strength workout steps

lriverawong:
## Summary

`create_strength_workout` fails for every exercise with `API Error 400 - Invalid category`.

`build_strength_json` hardcodes `"category": "UNASSIGNED"` on each step. Garmin rejects that value, so the tool is non-functional end to end — a bare `"Squat"` fails the same as any other name. It surfaces as *"Garmin Connect is unreachable"*, because the `except Exception` swallows the real cause.

The field came in with #197 (merged via the #201 batch) to *"satisfy Garmin Connect strict validation for strength exercises"*. That premise turns out to be wrong — Garmin does not require the field. Before #197 this builder sent no `category` key at all, and that payload uploads fine.

This omits `category` unless the caller supplies one, and passes a supplied value through uppercased instead of deriving it from the exercise name.

## Why not derive the category from the name

Garmin publishes its exercise taxonomy unauthenticated at [`Exercises.json`](https://connect.garmin.com/web-data/exercises/Exercises.json). It has **47 categories**, and `UNASSIGNED` and `OTHER` are not among them — which is exactly why they 400.

The 11-value list in the `add_workout_step` docstring (`workouts.py`) and `workout_templates.py` is an "e.g." example, not the real enum. It is missing `CARRY`, `SLED`, `SANDBAG`, `PUSH_UP`, `OLYMPIC_LIFT`, `LEG_CURL`, `HYPEREXTENSION`, `PLYO` and others. Anything that maps names onto that subset produces wrong-but-plausible values — a farmer's carry filed as `CARDIO` when `CARRY` exists.

That matters beyond tidiness: every exercise in the catalog carries `primaryMuscles`/`secondaryMuscles`, so `category` drives muscle attribution. A guessed category writes a silently wrong record; an omitted one is the honest encoding of "unknown". Since the caller of an MCP tool is usually an LLM, it is better placed to classify a lift than a keyword table in this library would be.

I deliberately did not vendor the 47 keys either — #208 already loads this same catalog at runtime behind an `lru_cache`, and a second frozen copy here would be one more list to drift. Once #208 lands, `create_strength_workout` could validate through its loader.

## Evidence

Probed against the live API; every workout deleted immediately after.

| step payload | result |
|---|---|
| no `category` key | accepted |
| `category="SQUAT"` | accepted |
| `category="CARRY"` / `"SLED"` / `"PUSH_UP"` / `"LEG_CURL"` | accepted |
| `category="UNASSIGNED"` | **400 Invalid category** |
| `category="OTHER"` | **400 Invalid category** |

The last four accepted values are all outside the 11-value list, which is the concrete demonstration that it is not the real enum.

Reading workouts back also showed that **`exerciseName` is not free text**, contrary to what the docstrings claim. Garmin accepts any value without error but only retains ones matching its own exercise keys — `BARBELL_BACK_SQUAT` and `FARMERS_CARRY` round-trip, `"My Weird Lift"` comes back as `""`. The exercise name survives in the step `description`. The README and both docstrings are corrected to say that rather than promising a passthrough that does not happen.

## Test plan

New tests appended at end-of-file in `tests/unit/test_workout_builders.py`, clear of #202's hunk in the same file:

- [x] no `category` key is emitted when the caller does not supply one — the regression guard that would have caught this
- [x] a supplied `category` is uppercased and passed through; other exercises stay untouched
- [x] an empty or non-string `category` raises `ValueError` rather than reaching the API
- [x] the exercise name lands in the step description

Removed the assertion that unknown names fall back to `Other`; `OTHER` is rejected, so that behaviour was never reachable.

`uv run pytest tests/unit tests/integration -q` → 408 passed. Also ran the real tool end to end against my account and deleted the result.

## Note on #196

#196 was closed by #201, but only its end-condition half was fixed. It also asked for the category to be populated from a real name→category mapping, and #197 answered that with a constant the API rejects. This does not restore that mapping — it makes `category` caller-supplied instead, for the reasons above. Worth reopening #196 or closing it against this.

Might also be worth replacing the "e.g." category lists in `workouts.py` and `workout_templates.py` with a link to `Exercises.json`; they have now been read as authoritative more than once. Left out to keep this scoped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,16 @@ Returns: `{"status": "success", "workout_id": 1234567890, ...}`
 
 ### `create_strength_workout`
 
-Creates a strength workout from a list of exercises. Unknown names fall back to a generic step with the original name preserved.
+Creates a strength workout from a list of exercises. Each becomes a reps-based step, with the
+name kept in the step description. The name is also sent as `exerciseName`, but Garmin only
+retains that when it matches one of its own exercise keys (e.g. `FARMERS_CARRY`) — any other
+value is accepted and then stored empty.
+
+`category` is optional and passed straight through. Omit it and the key is left out of the
+payload entirely, which Garmin accepts. Supply it and it must be one of Garmin's exercise
+categories — anything else, including `OTHER` and `UNASSIGNED`, is rejected with
+`400 - Invalid category`. The full list is published at
+[`Exercises.json`](https://connect.garmin.com/web-data/exercises/Exercises.json).
 
 ```json
 {
@@ -146,7 +155,8 @@ Creates a strength workout from a list of exercises. Unknown names fall back to 
   "exercises": [
     {"name": "Sentadillas", "sets": 3, "reps": 12, "rest_seconds": 90},
     {"name": "Flexiones",   "sets": 3, "reps": 15, "rest_seconds": 60},
-    {"name": "Peso muerto", "sets": 3, "reps": 10, "rest_seconds": 90}
+    {"name": "Peso muerto", "sets": 3, "reps": 10, "rest_seconds": 90},
+    {"name": "Farmers Carry 40m", "sets": 3, "reps": 1, "rest_seconds": 90, "category": "CARRY"}
   ]
 }
 ```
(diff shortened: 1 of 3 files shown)


Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #222. Thank you for the contribution @lriverawong!

PR #215, opened 2026-07-27
fix: map virtual challenge fields

### LINKED ISSUE #212: get_inprogress_virtual_challenges returns null for name/dates on real expedition data
get_inprogress_virtual_challenges correctly returns the right count of in-progress virtual challenges/expeditions (2, matching what's shown in the Garmin Connect app under Explore > Expeditions), but every entry comes back with name, start_date, and end_date all null — only the uuid is populated.

Looking at src/garmin_mcp/challenges.py, the parser reads:

"name": challenge.get("name") or challenge.get("challengeName")
"start_date": _parse_iso_date(challenge.get("startDate"))
"end_date": _parse_iso_date(challenge.get("endDate"))

That mapping looks like it was written for the badge-challenge shape (which works fine via get_non_completed_badge_challenges — name/startDate/endDate/progress_percent all come through there). The expedition/virtual-challenge endpoint appears to return a different JSON shape, so none of those keys match and everything falls through to null.

Repro:

get_inprogress_virtual_challenges(start=1, limit=20)
# -> {"total": 2, "challenges": [{"name": null, "uuid": "...", "start_date": null, "end_date": null}, ...]}

Ask: Could you log the raw response from the underlying garminconnect call for this endpoint and add the correct field names (likely something like expeditionName/virtualChallengeName and a percentComplete-style field)? Happy to test against a real account if useful.

pn:
## Summary
- map the live Garmin expedition fields badgeChallengeName, badgeProgressValue, and badgeTargetValue
- preserve both legacy field families and legitimate zero-valued progress
- format elevation, step, and time challenges with their actual units instead of labeling them as distance
- strengthen integration coverage with the payload shape observed from an active expedition

## Root cause
The virtual challenge curator expected generic challenge fields, but Garmin returns the badge-challenge field shape for active expeditions. This caused the name and all progress fields to be discarded. startDate and endDate remain null because Garmin itself returns null for non-expiring expeditions.

## Validation
- reproduced against an active expedition on a device-synced Garmin account
- verified that the patched mapper returns the expedition name and progress from the same live endpoint
- covered distance, elevation, steps, time, zero progress, zero target, and both legacy field families
- ran the full CI test selection: 411 passed

Fixes #212

### DIFF
--- a/src/garmin_mcp/challenges.py
+++ b/src/garmin_mcp/challenges.py
@@ -134,8 +134,8 @@ def _parse_iso_date(iso_string: str) -> str:
     return iso_string.split("T")[0] if "T" in iso_string else iso_string
 
 
-def _first_non_none(data: dict, *keys: str):
-    """Return the first present, non-None value for the supplied keys."""
+def _first_non_none(data: dict, *keys: str) -> Any:
+    """Return the first non-None value for the supplied keys."""
     for key in keys:
         value = data.get(key)
         if value is not None:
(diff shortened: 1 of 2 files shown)


Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #219. Thank you for the contribution @pn!

PR #214, opened 2026-07-26
feat: bump garminconnect to 0.3.7 and expose update_workout + get_exercise_types

tamcore:
python-garminconnect 0.3.7 brings a native update_workout method and a strength-exercise catalog. 

- update_workout(workout_id, workout_data): edit a workout in place via PUT, keeping its ID (schedules stay valid); reuses upload_workout's validators, prefers the server-returned name/id.
- get_exercise_types(category=None): list the strength-exercise catalog (categories + counts; per-category exercise + display names) from garminconnect.exercises, so strength steps use valid category / exerciseName keys.

python-garminconnect 0.3.7 requires Python >=3.12, so requires-python and the CI matrix are bumped accordingly (drops 3.10/3.11).

PR #216, opened 2026-07-28
fix: preserve misplaced workout target bounds

### LINKED ISSUE #210: Running pace.zone targetValueOne/Two accepted on upload but show as 0:00 on device and in preview

pn:
## Summary

Fix #210 by repairing workout target bounds that are mistakenly nested inside
`targetType`.

Garmin accepts that malformed payload without an API error and preserves
`pace.zone`, but silently discards the bounds. The resulting workout can still
show descriptive text such as `8:00–8:30/km`, while the device has no
machine-readable target range to track.

This change:

- moves nested bounds and zone numbers from `targetType` and
  `secondaryTargetType` to their canonical step-level positions;
- applies the repair recursively inside repeat groups;
- preflights the complete workout before mutation and reports both conflicting
  values with an actionable path;
- rejects targets that mix a named zone with a custom range instead of relying
  on Garmin's silent zone precedence;
- covers `upload_workout`, `upload_workouts`, and inline `schedule_workouts`;
- documents the correct sibling-field shape and the m/s units used by custom
  running pace ranges.

## Root cause and reproduction

The issue body originally showed the canonical shape:

```json
{
  "targetType": {
    "workoutTargetTypeId": 6,
    "workoutTargetTypeKey": "pace.zone"
  },
  "targetValueOne": 2.0833333,
  "targetValueTwo": 1.9607843
}
```

The later complete payloads revealed that the actual uploaded shape was:

```json
{
  "targetType": {
    "workoutTargetTypeId": 6,
    "workoutTargetTypeKey": "pace.zone",
    "targetValueOne": 2.0833333,
    "targetValueTwo": 1.9607843
  }
}
```

A live API and Forerunner 255 A/B test reproduced the difference:

| Case | Stored Garmin JSON | FIT export | Device |
|---|---|---|---|
| Bounds nested inside `targetType` | both bounds `null` | no target fields | preview says no target; descriptive screen remains, but no range tracking |
| Bounds beside `targetType` | both bounds preserved | `target_type=speed`, range `1.961–2.083 m/s` | preview shows `8:00–8:30/km`; range tracking works |

This rules out repeat-group nesting, bound ordering, and device-family behavior
as the cause of the reported symptom.

## Safety behavior

Unambiguous malformed input is repaired:

```json
{
  "targetType": {
    "workoutTargetTypeId": 6,
    "workoutTargetTypeKey": "pace.zone",
    "targetValueOne": 2.0833333
  }
}
```

becomes:

```json
{
  "targetType": {
    "workoutTargetTypeId": 6,
    "workoutTargetTypeKey": "pace.zone"
  },
  "targetValueOne": 2.0833333
}
```

If both positions are present with different values, the request is rejected
before upload with the exact workout-step path and both values. The complete
workout is preflighted before nested target fields are moved, so a conflict in a
later step cannot leave an earlier step partially normalized.

Named zones and custom ranges are mutually exclusive. Live API verification
showed that Garmin keeps `zoneNumber` and silently discards coexisting custom
bounds; the tools reject that ambiguous shape instead.

## Verification

- focused workout integration tests: `82 passed`
- Python 3.10.20: `422 passed`
- Python 3.11.15: `422 passed`
- Python 3.12.13: `422 passed`
- Python 3.13.13: `422 passed`
- live API probes confirmed:
  - nested primary `zoneNumber` is discarded while its step-level form survives;
  - a named zone takes precedence and causes coexisting custom bounds to be discarded;
  - nested secondary bounds are discarded while their step-level forms survive;
- live E2E on the exact inline `schedule_workouts` path:
  - malformed nested input sent through the MCP;
  - stored Garmin JSON preserved both repaired bounds;
  - downloaded FIT contained the intended speed range;
- `uv lock --check`
- `git diff --check`

## Related open work

No currently open PR overlaps this fix.

#214 adds an `update_workout` tool but is still open and not part of current
`main`. If #214 merges first, this branch should be rebased and that new
caller-supplied update path should invoke the same normalization helper before
validation.

Closes #210.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,9 @@ treats the numeric ID as authoritative: a mismatched payload such as
 `pace.zone`, because ID `6` means `pace.zone`.
 
 For a custom heart-rate range, use target type ID `4` with `heart.rate.zone` and
-put the bpm range in `targetValueOne` / `targetValueTwo`:
+put the bpm range in `targetValueOne` / `targetValueTwo`. These value fields
+belong on the workout step, alongside `targetType`; do not nest them inside the
+`targetType` object:
 
 ```json
 {
@@ -251,6 +253,25 @@ put the bpm range in `targetValueOne` / `targetValueTwo`:
   "targetValueTwo": 157
 }
 ```
+
+The same shape applies to a custom running pace range. Pace bounds use meters
+per second:
+
+```json
+{
+  "targetType": {
+    "workoutTargetTypeId": 6,
+    "workoutTargetTypeKey": "pace.zone"
+  },
+  "targetValueOne": 2.0833333,
+  "targetValueTwo": 1.9607843
+}
+```
+
+That example represents `8:00–8:30 min/km`. Garmin silently discards values
+nested inside `targetType`, leaving a pace target with no active range. The
+upload tools repair that unambiguous nesting mistake, but reject the request if
+nested and step-level values conflict.
 
 For a named Garmin HR zone, use the same target type with `zoneNumber` instead:
 
--- a/README.md
+++ b/README.md
@@ -263,15 +263,16 @@ per second:
     "workoutTargetTypeId": 6,
     "workoutTargetTypeKey": "pace.zone"
   },
-  "targetValueOne": 2.0833333,
-  "targetValueTwo": 1.9607843
+  "targetValueOne": 1.9607843,
+  "targetValueTwo": 2.0833333
 }
 ```
 
-That example represents `8:00–8:30 min/km`. Garmin silently discards values
-nested inside `targetType`, leaving a pace target with no active range. The
-upload tools repair that unambiguous nesting mistake, but reject the request if
-nested and step-level values conflict.
+That example represents `8:00–8:30 min/km`. The lower numeric bound is listed
+first for consistency with the heart-rate example; Garmin normalizes either
+bound order. Garmin silently discards values nested inside `targetType`, leaving
+a pace target with no active range. The upload tools repair that unambiguous
+nesting mistake, but reject the request if nested and step-level values conflict.
 
 For a named Garmin HR zone, use the same target type with `zoneNumber` instead:
 
@@ -284,6 +285,11 @@ For a named Garmin HR zone, use the same target type with `zoneNumber` instead:
   "zoneNumber": 3
 }
 ```
+
+Use either `zoneNumber` or `targetValueOne` / `targetValueTwo` on a target, not
+both. Garmin treats the named zone as authoritative and silently discards a
+coexisting custom range, so the upload tools reject that ambiguous shape.
+
 ## One-click Install (Claude Desktop)
 
 The easiest way to add this server to Claude Desktop is via the `.dxt` Desktop Extension file — no JSON editing required.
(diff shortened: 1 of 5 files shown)


Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #221. Thank you for the contribution @pn!

Issue #204, opened 2026-07-22
Bug: Nested variable (`${HOME}`) inside a `user_config` default is not resolved


gabrielecandela:
### Summary
When a `user_config` field's `default` value itself contains a Claude Desktop
template variable (e.g. `${HOME}`), that inner variable is not substituted.
The literal, unresolved string gets passed through to the server's
environment, causing Garmin login to fail even when valid tokens already
exist on disk.

### Environment
- Extension version: garmin-mcp 0.1.0 (.dxt)
- OS: macOS (darwin)
- Claude Desktop version: [fill in — Claude menu → About Claude Desktop]

### Steps to reproduce
1. Run `garmin-mcp-auth` in a terminal, authenticate successfully. Tokens
   are saved to `~/.garminconnect` (confirmed valid via
   `garmin-mcp-auth --verify`).
2. Install `garmin-mcp.dxt` in Claude Desktop.
3. Leave the "Token storage directory" field at its default
   (`${HOME}/.garminconnect`) — i.e. don't manually edit it.
4. Start a new chat and trigger any Garmin tool.

### Expected behavior
`GARMINTOKENS` env var resolves to `/Users/<me>/.garminconnect`, matching
where the tokens actually live, and the server logs in using the saved
token — no interactive login needed.

### Actual behavior
Server log shows the *literal, unresolved* string:

Trying to login to Garmin Connect using token data from directory '${HOME}/.garminconnect'...
Login tokens not present, login with your Garmin Connect credentials to generate them.

Since `${HOME}` was never expanded, the path doesn't exist, so the server
falls back to an interactive login with blank credentials, which then also
triggers Garmin's `429` IP rate limit on repeated attempts.

### Suspected root cause
`manifest.json` has:

```json
"env": { "GARMINTOKENS": "${user_config.token_path}" },
"user_config": {
  "token_path": { "default": "${HOME}/.garminconnect" }
}
```

This requires two passes of substitution: `${user_config.token_path}` →
default value → then `${HOME}` inside *that* value. It looks like Claude
Desktop only performs one pass — it swaps in the field's default string but
doesn't re-scan the result for further template variables. So any
`user_config` default that itself references a built-in variable (`${HOME}`,
`${DESKTOP}`, etc.) silently fails to resolve, but only when the user
leaves the field untouched (manually typing the resolved path works fine,
since that skips the nested reference entirely).

### Workaround
Manually enter the fully resolved path (e.g.
[...]

PR #217, opened 2026-07-28
fix(dxt): resolve default Garmin token path

pn:
## Summary

Fix the Desktop Extension interpolation failures reported in #204.

- change the `user_config.token_path` default from `${HOME}/.garminconnect` to `~/.garminconnect`, avoiding nested template interpolation in Claude Desktop
- centralize token-path normalization in `token_utils` for both the MCP server and `garmin-mcp-auth`
- expand environment variables, `~`, and a literal `${HOME}` fallback consistently for token and base64-token paths
- treat the exact unresolved optional email/password placeholders as unset, preventing a non-interactive credential-login attempt and potential Garmin `429`
- rebuild the committed `.dxt` from a clean archive and add regression coverage for manifest/bundle drift

## Compatibility rationale

The manifest change prevents new installations from relying on nested interpolation. Runtime normalization remains a compatibility fallback for existing manifests, custom MCP configurations, and explicit CLI paths that still contain `${HOME}`. On Windows, the explicit fallback resolves the home directory through Python even when `HOME` is unset and `USERPROFILE` is the available source.

Credential handling only rejects the exact `${user_config.garmin_email}` and `${user_config.garmin_password}` placeholders. Real credential values containing `$` or other template-like text remain unchanged.

Claude Desktop may display the literal default `~/.garminconnect` in its directory field. The server resolves it to the platform home directory before any token access.

## Bundled manifest correction

The committed bundle on `main` was stale relative to `dxt/manifest.json`. Rebuilding it here also includes the already-merged #203 correction from `server.type: uv` to `server.type: python`. This is intentional: the source manifest and shipped bundle now match, and a regression test enforces that relationship.

## Delivery requirement

The published `0.1.0` release asset is unchanged by merging this PR. A maintainer must bump the extension manifest version and publish a new release containing the rebuilt `garmin-mcp.dxt` before users following the README can distinguish and receive the fix.

This PR intentionally does not choose the next release version or introduce a release workflow because the repository has no established automated release/versioning policy. Those are separate maintainer decisions. #204 should remain open until a fixed release asset is published.

## Verification

- `uv run pytest tests/unit tests/integration -q` — 419 passed
- focused authentication, user-config, token-path, and DXT tests — 66 passed
- `git diff --check`
- rebuilt archive passes `unzip -t garmin-mcp.dxt`
- regression coverage exercises unresolved optional credentials, the `HOME`-unset fallback, absolute-path idempotence, base64-token paths, CLI write/read consistency, and bundle/source equality

Manual Claude Desktop installation with the directory field left at its default was not performed. That end-to-end check remains required before publishing the release asset.

Addresses #204.


Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #220. Thank you for the contribution @pn!

PR #227, opened 2026-07-28
fix: cap mcp <2 so fresh installs stop resolving to the 2.x API

### LINKED ISSUE #226: Fresh install fails with MCP 2.0 (ModuleNotFoundError: mcp.server.fastmcp)
A fresh installation currently fails because the project imports the v1 MCP API:

```python
from mcp.server.fastmcp import FastMCP
```

However, `pyproject.toml` specifies:

```toml
mcp>=1.28.1
```

Since MCP 2.0 has been released, `uv` resolves the dependency to the latest 2.x version, where `mcp.server.fastmcp` no longer exists.

## Reproduction

```bash
uvx --refresh \
  --python 3.12 \
  --from git+https://github.com/Taxuspt/garmin_mcp \
  garmin-mcp-auth
```

## Actual Behavior

The command fails immediately with:

```text
Traceback (most recent call last):
  File ".../bin/garmin-mcp-auth", line 6, in <module>
    from garmin_mcp.auth_cli import main
  File ".../site-packages/garmin_mcp/__init__.py", line 10, in <module>
    from mcp.server.fastmcp import FastMCP
```

## Expected Behavior

`garmin-mcp-auth` should launch successfully on a fresh installation.

## Suggested Fix

Until the project is migrated to the MCP 2.x API, consider pinning the MCP dependency to the latest v1 release, for example:

```toml
mcp>=1.28.1,<2
```

or another appropriate v1 constraint.

This should prevent fresh installations from resolving to the incompatible MCP 2.x release while allowing time for a migration.

phdelodder:
Fixes #226.

`mcp` 2.0.0 is now on PyPI and renamed the `mcp.server.fastmcp` package to `mcp.server.mcpserver` (`FastMCP` -> `MCPServer`). Because the dependency is declared as an open-ended `mcp>=1.28.1`, every fresh resolve picks up 2.x and the server dies at import:

```
```

This hits any `uvx --from git+...` launch, which re-resolves on each run — including the DXT / Claude Desktop entry point — so installs that worked yesterday break with no local change and no obvious cause (Claude Desktop only reports a generic disconnect).

This is the fix suggested in #226: cap at `<2` until the server is ported to the `MCPServer` API. `uv.lock` already resolves mcp 1.28.1, so only the requirement string changes there.

## Verification

- `uvx --refresh --from . --python 3.12 garmin-mcp` starts again (fails on `main` with the traceback above).
- 455 integration + unit tests pass.

Note this is a stopgap — the real follow-up is porting `mcp.server.fastmcp.FastMCP` to `mcp.server.mcpserver.MCPServer`, which I have not attempted here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,9 @@ dependencies = [
     "python-dotenv==1.2.2",
     "garminconnect==0.3.2",
     "requests==2.33.0",
-    "mcp>=1.28.1",
+    # mcp 2.x renamed mcp.server.fastmcp -> mcp.server.mcpserver (FastMCP ->
+    # MCPServer). Cap until the server is ported to the new API.
+    "mcp>=1.28.1,<2",
     "fitparse>=1.2.0",
 ]
 
--- a/uv.lock
+++ b/uv.lock
@@ -358,7 +358,7 @@ dev = [
 requires-dist = [
     { name = "fitparse", specifier = ">=1.2.0" },
     { name = "garminconnect", specifier = "==0.3.2" },
-    { name = "mcp", specifier = ">=1.28.1" },
+    { name = "mcp", specifier = ">=1.28.1,<2" },
     { name = "python-dotenv", specifier = "==1.2.2" },
     { name = "requests", specifier = "==2.33.0" },
 ]


pn (approved):
Approved. I reproduced the failure with MCP 2.0.0 and verified that the `<2` constraint restores the affected fresh-install and authentication path. The change is focused, the lockfile metadata is consistent, and all checks pass.

The MCP 2 migration can be handled separately and should not block this hotfix.

pn:
@Taxuspt, could you please consider merging this hotfix soon?

This is effectively an availability regression for every fresh or dynamic `uvx` resolution, including the documented authentication command, direct GitHub client configurations, and the Claude Desktop/DXT path. The failure happens during import, before the server or authentication CLI can start; GUI clients may surface it only as a generic disconnect. Because `uvx` can re-resolve dependencies, users can also break without changing Garmin MCP itself.

The merge risk appears low:

- the change is a single, focused dependency upper bound;
- the PR is currently clean and all reported CI, validation, installation, security, and Python-version checks pass;
- the MCP 2.0 failure and the v1 constraint were independently reproduced and verified in review;
- the official MCP Python SDK guidance explicitly recommends a `<2` upper bound for projects that have not migrated;
- the checked-in lockfile already uses MCP 1.28.1, so this aligns package metadata with the version already tested by the repository.

A full migration from `FastMCP` to the v2 `MCPServer` API is worthwhile, but it is a broader breaking-change migration and can follow separately. Keeping it separate should let this small, reversible compatibility fix restore installs immediately.

curry684 (approved):

Stephan202 (approved):

PR #228, opened 2026-07-28
feat(training): fall back to profile VO2 max when trend history is unavailable

guarismo:
## Summary

`get_vo2max_trend` currently only reads `get_training_status(date)` per day and returns "No VO2 max data found" if that endpoint has no data across the whole range. In practice, some accounts/devices don't populate `mostRecentVO2Max` on `get_training_status` for a given day even though a current estimate is available elsewhere.

This PR makes the lookup more tolerant:

- Per day, tries `get_max_metrics`, `get_fitnessage_data`, then `get_training_status` in turn (first one with a usable value wins), via a small `_extract_vo2_value` helper that checks several known Garmin response shapes (`vo2MaxValue`, `vo2MaxPreciseValue`, nested `mostRecentVO2Max.generic`/`.cycling`, `userData.vo2Max*`, etc.).
- If no historical data point is found anywhere in the requested range, falls back to the current estimate from `get_user_profile()` and adds a `note` field explaining that it's a current snapshot rather than historical trend data, instead of just returning a "no data found" string.
- Each trend point now also records its `source` method, for debuggability.

## Test plan

- [x] `test_get_vo2max_trend_falls_back_to_profile` — verifies the profile fallback path and its `note`.
- [x] `test_get_vo2max_trend_uses_daily_metrics` — verifies `get_max_metrics` is used when it returns data, and `source` is recorded.
- [x] Ran the full existing test suite locally; no other tests affected.

### DIFF
--- a/src/garmin_mcp/training.py
+++ b/src/garmin_mcp/training.py
@@ -4,7 +4,7 @@ Training and performance functions for Garmin Connect MCP Server
 
 import json
 import datetime
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 # The garmin_client will be set by the main file
 garmin_client = None
@@ -25,29 +25,41 @@ def _as_dict(value: Any) -> Dict[str, Any]:
     return value if isinstance(value, dict) else {}
 
 
-def _extract_vo2_value(data: Any) -> Optional[float]:
-    """Find a VO2 max value in known Garmin response shapes."""
+def _extract_vo2_measurement(data: Any) -> Optional[Tuple[float, str]]:
+    """Find a VO2 max value and sport in known Garmin response shapes."""
+    if isinstance(data, list):
+        for item in data:
+            measurement = _extract_vo2_measurement(item)
+            if measurement is not None:
+                return measurement
+        return None
+
     data = _as_dict(data)
+    # Garmin uses "generic" for its running/non-cycling VO2 max series.
     candidate_paths = (
-        ("vo2MaxRunning",),
-        ("vo2MaxCycling",),
-        ("vo2Max",),
-        ("vo2MaxValue",),
-        ("vo2MaxPreciseValue",),
-        ("mostRecentVO2Max", "generic", "vo2MaxValue"),
-        ("mostRecentVO2Max", "generic", "vo2MaxPreciseValue"),
-        ("mostRecentVO2Max", "cycling", "vo2MaxValue"),
-        ("mostRecentVO2Max", "cycling", "vo2MaxPreciseValue"),
-        ("userData", "vo2MaxRunning"),
-        ("userData", "vo2MaxCycling"),
+        (("vo2MaxRunning",), "running"),
+        (("vo2MaxCycling",), "cycling"),
+        (("vo2Max",), "running"),
+        (("vo2MaxValue",), "running"),
+        (("vo2MaxPreciseValue",), "running"),
+        (("generic", "vo2MaxValue"), "running"),
+        (("generic", "vo2MaxPreciseValue"), "running"),
+        (("cycling", "vo2MaxValue"), "cycling"),
+        (("cycling", "vo2MaxPreciseValue"), "cycling"),
+        (("mostRecentVO2Max", "generic", "vo2MaxValue"), "running"),
+        (("mostRecentVO2Max", "generic", "vo2MaxPreciseValue"), "running"),
+        (("mostRecentVO2Max", "cycling", "vo2MaxValue"), "cycling"),
+        (("mostRecentVO2Max", "cycling", "vo2MaxPreciseValue"), "cycling"),
+        (("userData", "vo2MaxRunning"), "running"),
+        (("userData", "vo2MaxCycling"), "cycling"),
     )
 
-    for path in candidate_paths:
+    for path, sport in candidate_paths:
         current: Any = data
         for key in path:
             current = _as_dict(current).get(key)
-        if isinstance(current, (int, float)):
-            return float(current)
+        if isinstance(current, (int, float)) and not isinstance(current, bool):
+            return float(current), sport
     return None
 
 
@@ -1025,6 +1037,9 @@ def register_tools(app):
         Note: VO2 max estimates are smoothed and update gradually — daily changes of <0.5
         are within normal noise. Focus on the 4-6 week trend direction.
 
+        If historical values are unavailable, the current profile estimate is returned
+        separately and is not represented as a historical trend point.
+
         Recommended range: 4-12 weeks. Maximum: 90 days.
 
         Args:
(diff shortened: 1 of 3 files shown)


src/garmin_mcp/training.py
```diff
@@ -20,6 +20,37 @@ def configure(client):
...
 
 
+def _as_dict(value: Any) -> Dict[str, Any]:
+    """Return a dict for Garmin sections that may be null or another shape."""
+    return value if isinstance(value, dict) else {}
+
+
+def _extract_vo2_value(data: Any) -> Optional[float]:
+    """Find a VO2 max value in known Garmin response shapes."""
+    data = _as_dict(data)
```
pn, line 38: Thanks for adding a shared extractor here. Could we normalize the actual `get_max_metrics` payload before coercing it to a dict? That endpoint commonly returns a list whose item stores the value under `generic.vo2MaxValue` or `generic.vo2MaxPreciseValue`; the current `_as_dict` call discards the list, and the top-level `generic` paths are not included either. The observed `get_fitnessage_data` schema contains fitness-age fields and components but no supported top-level VO₂ field, so I would also remove that source unless a captured response demonstrates one. Please consider updating the test to use the representative list-of-dicts max-metrics shape rather than the flat `{"vo2MaxValue": 27.5}` mock.

src/garmin_mcp/training.py
```diff
@@ -1018,36 +1049,76 @@ async def get_vo2max_trend(start_date: str, end_date: str) -> str:
...
-                        vo2_rounded = round(vo2, 1)
-                        if vo2_rounded != last_vo2:  # deduplicate unchanged values
-                            trend.append({"date": date_str, "vo2_max": vo2_rounded})
-                            last_vo2 = vo2_rounded
-            except Exception:
-                pass
+            vo2 = None
+            source = None
+
+            for method_name in (
```
pn: Could we keep the request fan-out bounded here? For an empty or sparse range this currently makes three sequential Garmin requests per day before the profile fallback—up to 270 daily requests for the allowed 90-day range, versus 90 previously. Since the server already treats Garmin throttling as a live failure mode, putting `get_training_status` first, removing the ineffective fitness-age lookup, or otherwise avoiding all three calls for every missing day would make this safer while preserving the fallback.

src/garmin_mcp/training.py
```diff
@@ -1018,36 +1049,76 @@ async def get_vo2max_trend(start_date: str, end_date: str) -> str:
...
+            try:
+                profile = garmin_client.get_user_profile()
+            except Exception:
+                profile = None
+
+            current_vo2 = _extract_vo2_value(profile)
+            if current_vo2 is None:
+                return f"No VO2 max data found between {start_date} and {end_date}."
+
+            trend.append(
```
pn: Could we keep the current profile estimate separate from the historical `trend`? For example, a request for a 2023 range currently records today's profile value using the requested `end_date`, reports `data_points: 1`, and calculates `change: 0.0`. Those machine-readable fields make the value look like a real observation from that period despite the explanatory note. I suggest returning an empty historical trend with null historical summaries, and exposing the fallback under a separate field such as `current_vo2_max_estimate` with an explicit current or unknown `as_of` date.

src/garmin_mcp/training.py
```diff
@@ -20,6 +20,37 @@ def configure(client):
...
+    data = _as_dict(data)
+    candidate_paths = (
+        ("vo2MaxRunning",),
+        ("vo2MaxCycling",),
+        ("vo2Max",),
+        ("vo2MaxValue",),
+        ("vo2MaxPreciseValue",),
+        ("mostRecentVO2Max", "generic", "vo2MaxValue"),
+        ("mostRecentVO2Max", "generic", "vo2MaxPreciseValue"),
+        ("mostRecentVO2Max", "cycling", "vo2MaxValue"),
```
pn: Could the extractor also preserve which sport produced the value? As written, one day can select the generic/running value and another day can fall through to cycling, after which the tool reports their difference as a single VO₂ trend while both points only identify the API method as their source. Returning the sport with the value and keeping one sport consistent across the range—or exposing separate running and cycling series—would avoid comparing different modalities.

pn (commented):
Thanks for working on a more resilient VO₂ max fallback. I reviewed this against the pinned garminconnect 0.3.2 behavior and left a few inline suggestions around response shapes, request volume, and preserving the meaning of historical trend data.

guarismo:
> Thanks for working on a more resilient VO₂ max fallback. I reviewed this against the pinned garminconnect 0.3.2 behavior and left a few inline suggestions around response shapes, request volume, and preserving the meaning of historical trend data.

I will take a look.
Honestly, I'm not familiar with the API, I was mostly guessing with Claude help.

pn:
Hi @guarismo — thank you for the work on this VO₂ max fallback.

Following the review, I prepared a follow-up PR against your `pr/vo2-trend-fallback` branch rather than opening a competing PR to `main`:

https://github.com/guarismo/garmin_mcp/pull/1

The follow-up corrects parsing for Garmin’s actual list and nested max-metrics responses, preserves running and cycling measurements from mixed payloads, keeps profile estimates separate from historical trends, and removes fitness-age as a fallback because that response does not contain a VO₂ value.

It also fetches max metrics with one date-range request and queries training status only for uncovered dates. Against the original PR head, this reduces the 90-day no-history path from 271 requests to 92 when range support is available.

For covered dates, per-day max-metrics values intentionally take precedence over the carried-forward training-status value. This can produce more accurate change dates and a different number of trend points for users whose trend previously came from training status.

The two existing VO₂ test names are preserved, with their fixtures updated to Garmin’s real payload shapes. Coverage has grown to 11 integration tests and 7 unit-test functions, and the full unit and integration suite passes: **481 tests**.

I also performed read-only validation using the pinned `garminconnect==0.3.2`. Garmin accepted both 28-day and 90-day range requests, sampled range values matched their single-day responses, and sparse range coverage correctly fell back to training status. The request uses the client library’s existing endpoint constant and `connectapi` method.

The cycling regression from the review now preserves `55.0 → 56.0 → 57.0` even when later payloads contain both running and cycling measurements.

If the approach looks good to you, merging the follow-up into your branch should update this PR automatically. Please feel free to suggest changes or adopt only the parts you prefer.

src/garmin_mcp/training.py
pn: Resolved in the current head: running and cycling measurements are collected separately, and one sport is selected consistently for the trend.

src/garmin_mcp/training.py
pn: Resolved in the current head: max-metrics list/nested payloads are parsed, fitness-age is no longer a fallback source, and tests use representative response shapes.

src/garmin_mcp/training.py
pn: Resolved in the current head: the profile fallback is returned separately as `current_vo2_max_estimate`, with no fabricated historical point.

src/garmin_mcp/training.py
pn: Resolved in the current head: max metrics are fetched once for the range, with training status queried only for uncovered dates.

Taxuspt:
Closed: the changes from this PR were applied locally, tested with CI, and merged via PR #240. Thank you for the contribution @guarismo!

PR #233, opened 2026-07-31
fix: pin mcp SDK below 2.0 to keep FastMCP import working

GeorginaSteele:
## Problem

mcp 2.0.0 (released on PyPI 2026-07-28) removed `mcp.server.fastmcp` entirely, replacing `FastMCP` with a new `MCPServer` API. Since the dependency constraint is `mcp>=1.28.1`, any fresh install or uvx environment rebuild now resolves to 2.0.0 and the server crashes on startup:

```
```

Users with a uvx/uv cache built before July 28 are unaffected until their environment rebuilds (e.g. when a new commit lands on this repo), so this breaks installs progressively.

## Fix

Pin `mcp>=1.28.1,<2.0` until the codebase is migrated to the new `MCPServer` API. The 2.0 migration is a substantial rewrite (all tool/resource registration modules use the `FastMCP` decorator pattern), so pinning keeps installs working in the meantime.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "python-dotenv==1.2.2",
     "garminconnect==0.3.2",
     "requests==2.33.0",
-    "mcp>=1.28.1",
+    "mcp>=1.28.1,<2.0",
     "fitparse>=1.2.0",
 ]
 


pn (commented):
Thank you for taking the time to investigate this and submit a fix. This is already tracked in #226, and the earlier PR #227 implements the same `mcp>=1.28.1,<2` cap.

There is also a lockfile consistency issue here: this PR updates only `pyproject.toml`, while the dependency specifier is recorded in `uv.lock` as well. I verified that `uv lock --check` on this change reports that the lockfile needs to be updated. PR #227 updates both files, and the same check succeeds there.

To avoid duplicating work, could you please close this PR in favor of #227? Thanks again for contributing.

Taxuspt:
Closed: PR #227 fixes the same issue (mcp>=1.28.1,<2) and was already merged. Thank you for the contribution!

PR #234, opened 2026-08-01
fix: cap mcp dependency below 2.0 to keep FastMCP import working

guysherman:
## Summary

`garmin_mcp` fails to start at import time on a fresh install. `pyproject.toml` declares `mcp>=1.28.1` with no upper bound, so a fresh resolve now pulls `mcp==2.0.0`. That release removed the bundled FastMCP module (`mcp.server.fastmcp`), which this project imports in `garmin_mcp/__init__.py`. Result: the server process crashes immediately after the client sends `initialize`.

```
Traceback (most recent call last):
  File ".../bin/garmin-mcp", line 6, in <module>
    from garmin_mcp import main
  File ".../site-packages/garmin_mcp/__init__.py", line 10, in <module>
    from mcp.server.fastmcp import FastMCP
```

This is not a cache issue — `uvx --refresh` / `uv cache clean` don't help because the resolver still picks 2.0.0.

## Fix

Cap the `mcp` dependency at `<2` in `pyproject.toml` (and regenerate `uv.lock`), keeping the current FastMCP import working with no code changes. This unblocks users immediately; migrating to the standalone `fastmcp` package for `mcp` 2.x support is a separate, larger follow-up if maintainers want it.

## Test plan

- [x] `uv run python -c "from mcp.server.fastmcp import FastMCP"` succeeds with the capped dependency (previously failed against `mcp==2.0.0`)
- [x] `uv run python -c "import garmin_mcp"` succeeds
- [x] `uv run pytest -q` — 467 passed; the only failures are pre-existing live-API e2e tests requiring real Garmin credentials, unrelated to this change
- [x] `uvx --python 3.12 --from git+https://github.com/guysherman/garmin_mcp@fix/cap-mcp-dependency-below-2 garmin-mcp` installs and initializes successfully end-to-end (matches how the `.dxt` extension launches the server)
- [x] Built and installed the extension and it is working fine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "python-dotenv==1.2.2",
     "garminconnect==0.3.2",
     "requests==2.33.0",
-    "mcp>=1.28.1",
+    "mcp>=1.28.1,<2",
     "fitparse>=1.2.0",
 ]
 
--- a/uv.lock
+++ b/uv.lock
@@ -321,7 +321,7 @@ name = "exceptiongroup"
 version = "1.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
+    { name = "typing-extensions" },
 ]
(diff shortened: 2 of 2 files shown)


pn:
Thank you for taking the time to investigate this and for documenting the verification so thoroughly. This issue is already tracked in #226, and the earlier PR #227 implements the same `mcp>=1.28.1,<2` cap, including the corresponding `uv.lock` update.

To keep the discussion and review consolidated, could you please close this PR in favor of #227? Thanks again for contributing.

guysherman:
Closing as duplicate of #227

PR #237, opened 2026-08-02
implementation of fix which stopped server for initialising  

vedantwalia:
…is includes changes to the unit and e2e tests to reflect the new server parameters and startup process. The tests have been modified to ensure they are compatible with the updated server initialization logic.

I was unable go sync my Garmin post the MCP backend update so figured ill formalise the fix and help a bit

### DIFF
--- a/tests/e2e/test_server_e2e.py
+++ b/tests/e2e/test_server_e2e.py
@@ -13,6 +13,8 @@ Run with: pytest tests/e2e/ -m e2e
 Or skip with: pytest -m "not e2e"
 """
 
+import os
+import sys
 import pytest
 import asyncio
 from datetime import datetime
@@ -27,6 +29,22 @@ from mcp.client.stdio import stdio_client
 load_dotenv()
 
 
+def _build_server_params():
+    """Construct MCP server parameters using the active Python interpreter."""
+    env = os.environ.copy()
+    repo_root = Path(__file__).resolve().parents[2]
+    src_path = repo_root / "src"
+    if str(src_path) not in env.get("PYTHONPATH", ""):
+        env["PYTHONPATH"] = os.pathsep.join(
+            filter(None, [str(src_path), env.get("PYTHONPATH", "")])
+        )
+    return StdioServerParameters(
+        command=sys.executable,
+        args=["-m", "garmin_mcp"],
+        env=env,
+    )
+
+
 @pytest.mark.e2e
 @pytest.mark.asyncio
 @pytest.mark.timeout(30)  # Pytest timeout
@@ -39,11 +57,7 @@ async def test_mcp_server_connection():
     - May require MFA code input if tokens are expired
     """
     # Use python module execution instead of direct script path
-    server_params = StdioServerParameters(
-        command="python",
-        args=["-m", "garmin_mcp"],
-        env=None,  # Uses current environment which includes .env variables
-    )
+    server_params = _build_server_params()
 
     # Connect to server with timeout
     try:
@@ -77,11 +91,7 @@ async def test_mcp_server_connection():
 @pytest.mark.timeout(30)
 async def test_list_activities_tool():
     """Test the list_activities MCP tool with real API"""
-    server_params = StdioServerParameters(
-        command="python",
-        args=["-m", "garmin_mcp"],
-        env=None,
-    )
+    server_params = _build_server_params()
 
     try:
         async with asyncio.timeout(20):
@@ -111,11 +121,7 @@ async def test_list_activities_tool():
 @pytest.mark.timeout(30)
 async def test_get_steps_data_tool():
     """Test the get_steps_data MCP tool with real API"""
-    server_params = StdioServerParameters(
-        command="python",
-        args=["-m", "garmin_mcp"],
-        env=None,
-    )
+    server_params = _build_server_params()
 
     try:
         async with asyncio.timeout(20):
@@ -147,11 +153,7 @@ async def test_get_steps_data_tool():
 @pytest.mark.timeout(45)
 async def test_multiple_tools():
     """Test multiple MCP tools in a single session"""
-    server_params = StdioServerParameters(
-        command="python",
-        args=["-m", "garmin_mcp"],
-        env=None,
-    )
+    server_params = _build_server_params()
 
     try:
         async with asyncio.timeout(40):
@@ -202,11 +204,7 @@ async def test_schedule_workouts_tool():
     import os
     import json
 
-    server_params = StdioServerParameters(
-        command="python",
-        args=["-m", "garmin_mcp"],
-        env=None,
-    )
+    server_params = _build_server_params()
 
     workout_ids_env = os.environ.get("GARMIN_TEST_WORKOUT_IDS", "")
     dates_env = os.environ.get("GARMIN_TEST_SCHEDULE_DATES", "")
@@ -265,11 +263,7 @@ async def test_upload_workouts_tool():
     """
     import json
 
-    server_params = StdioServerParameters(
-        command="python",
-        args=["-m", "garmin_mcp"],
-        env=None,
-    )
+    server_params = _build_server_params()
 
     minimal_workout = {
         "workoutName": "e2e Test Workout - DELETE ME",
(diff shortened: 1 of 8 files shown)


pn (commented):
Thanks for working on this. I reviewed the PR head and reproduced the relevant checks locally. The e2e parameter changes look sound: `sys.executable` avoids starting the server with a different Python interpreter, while explicitly passing the environment ensures the Garmin variables reach the subprocess. With `env=None`, the MCP SDK only inherits a small allowlist of environment variables.

My main concern is the addition of `fastmcp>=2`. Nothing in `src/` or the tests imports the standalone `fastmcp` package; the repository uses `mcp.server.fastmcp`, which is already provided by the existing `mcp` dependency.

The lockfile change adds exactly 37 packages, resolves the constraint to FastMCP 3.4.5, and upgrades `uvicorn` from 0.34.0 to 0.52.1. I uninstalled `fastmcp` and `fastmcp-slim` from the PR environment and all 456 integration and unit tests still passed. The applicable transitive dependencies would also be installed in the Linux Docker image.

I’d therefore recommend removing `fastmcp>=2` and regenerating `uv.lock`.

I also noticed that the title describes an update to the server startup mechanism, while no production files under `src/` are changed. Could you clarify whether a runtime change is missing, or whether this PR is intended specifically as a test-harness fix? If it’s the latter, updating the title and description would make the scope clearer.

One optional improvement: the new startup test currently uses the private `_tool_manager._tools` field and only asserts that at least one tool exists. Using the public `list_tools()` API and checking a few expected tools—or a meaningful minimum count—would make the smoke test more robust.

I’m reviewing this as a community contributor rather than a maintainer, so I’ll leave the merge decision to the maintainers. From my review, removing the unused dependency and clarifying the scope are the main points to address.

vedantwalia:
I have done the suggested changes and have updated the title as suggested. I pushed this PR since the MCP server has been breaking since the MCP 2.0 release, as documented in issue #226

pn:
Thanks for the update. I need to correct one part of my earlier review.

The import analysis was correct, but my conclusion about the dependency was incomplete. This project does not import the standalone `fastmcp` package; however, the original `fastmcp==3.4.5` dependency graph, through `fastmcp-slim[client]`, constrains the official `mcp` package to `>=1.24,<2`. Therefore, the original pair:

```toml
mcp>=1.28.1
fastmcp>=2
```

indirectly kept `mcp` on 1.28.1 and could make issue #226 appear fixed. My recommendation to remove standalone FastMCP was still the cleaner direction, but I should have said that it must be replaced by an explicit direct constraint such as `mcp>=1.28.1,<2`. Sorry for missing that dependency-resolution effect.

The naming here is unusually confusing:

- Official MCP SDK v1: install `mcp`; this repository imports `mcp.server.fastmcp.FastMCP`.
- Standalone FastMCP: install `fastmcp`; applications using it normally import `fastmcp.FastMCP`.
- Official MCP SDK v2: still install `mcp`; the new server API is `MCPServer`, exported from `mcp.server`.

Thus, `fastmcp>=2` means version 2 or later of the separate standalone FastMCP project. It does not mean MCP SDK 2.0. This PR does not migrate the application to MCP SDK 2: no production files under `src/` change, the v1 `mcp.server.fastmcp` import remains, and there is no `MCPServer` usage.

On the current PR head, I reproduced the consequence of removing `fastmcp` without adding `mcp<2`: a fresh unlocked Python 3.12 install resolves MCP 2.x and `garmin-mcp-auth --help` fails with `ModuleNotFoundError: No module named mcp.server.fastmcp`. The lockfile still passes and all 456 locked integration/unit tests pass because it remains locked to MCP 1.28.1.

Since #227 already contains the minimal direct `mcp>=1.28.1,<2` fix for #226, my suggestion as a community reviewer is:

1. Let #227 own the runtime compatibility fix, or coordinate with it rather than duplicating the same pin.
2. Keep #237 focused on its useful test-harness work: using `sys.executable`, forwarding the subprocess environment, and strengthening the startup smoke test.
3. Update the title/body to reflect that narrower scope; the current title still implies an MCP server migration.
4. Restore unrelated lockfile churn, including the `uvicorn` 0.34.0 → 0.52.1 change, and remove the personal `.vscode/settings.json` file.

The startup-test feedback has been addressed well. The remaining issue is ensuring the dependency fix is not lost while the two PRs are coordinated.

vedantwalia:
I have implemented the suggested changes, apologies for the shoddy work, this is my first FOSS PR, have been using OSS for nearly a decade now, wanted to give back to the community, am learning new things every day

uv.lock
```diff
@@ -1106,14 +1108,14 @@ wheels = [
 
 [[package]]
 name = "uvicorn"
-version = "0.34.0"
+version = "0.52.1"
```
pn, line 1111: The `uvicorn` 0.34.0 → 0.52.1 upgrade—and the nearby `resolution-markers`/`exceptiongroup` churn—does not follow from adding `mcp<2`. The base lock already resolves MCP 1.x, so the dependency graph need not change; only the project `requires-dist` metadata needs the new constraint. Please restore the base lockfile and retain only that metadata update, or regenerate it with the same uv version that produced the base lock, so this PR does not silently upgrade unrelated runtime packages.

tests/unit/test_project_dependencies.py
```diff
@@ -0,0 +1,11 @@
...
+
+from pathlib import Path
+
+
+def test_project_caps_mcp_to_v1_series() -> None:
+    """The project should explicitly stay on the MCP v1 series for compatibility."""
+    repo_root = Path(__file__).resolve().parents[2]
+    pyproject_text = (repo_root / "pyproject.toml").read_text()
+
+    assert '"mcp>=1.28.1,<2"' in pyproject_text
```
pn, line 11: Non-blocking: this asserts an exact serialized TOML string, so harmless formatting or dependency reordering can break the test without changing the requirement. If you keep this regression test, parsing the TOML and checking the `mcp` specifier semantically would be sturdier. Because the project supports Python 3.10, that would need either a declared `tomli` fallback or another Python-3.10-compatible parser.

tests/e2e/test_server_e2e.py
```diff
@@ -202,11 +204,7 @@ async def test_schedule_workouts_tool():
     import os
```
pn, line 204: Non-blocking cleanup: `os` is now imported at module scope for `_build_server_params()`, so this local import is redundant.

pn (commented):
Thanks for the update—and no need to apologize. Iterating on review feedback is a normal part of contributing, and you addressed the core issue quickly.

I re-reviewed the unchanged head (`3c9aa11`) and independently verified:

- The direct `mcp>=1.28.1,<2` constraint fixes the fresh-install failure. `garmin-mcp-auth --help` succeeds from fresh unlocked installs on Python 3.10, 3.12, and 3.13.
- All 457 unit and integration tests pass on Python 3.10, 3.12, and 3.13, and `uv lock --check` passes.
- The startup smoke test now uses the public `list_tools()` API with meaningful assertions.
- The e2e harness now uses `sys.executable` and forwards the environment.
- The personal `.vscode/settings.json` file has been removed.

GitHub currently reports no checks for this branch, so the local matrix above is the available execution evidence.

I left one substantive inline cleanup request on `uv.lock`, plus two explicitly non-blocking nits. The lockfile still contains unrelated resolution churn, including the `uvicorn` upgrade. Since the base lock already resolves MCP 1.x, the compatibility constraint does not require changing the resolved package graph.

There is also unresolved overlap with #227, which implements the same `mcp<2` cap and remains open. I’m reviewing as a community contributor rather than a maintainer, so which PR should own that fix is for the maintainers to decide; please acknowledge the overlap in this PR’s description.

A clearer title would be:

`fix: cap MCP to v1 and harden server startup tests`

Production code still imports `mcp.server.fastmcp.FastMCP` and does not use the v2 `MCPServer` API, so this is a compatibility pin plus test-harness improvements, not an MCP SDK 2 migration.

Functionally, the original installation problem is fixed. The unrelated lockfile changes are the only concrete code-cleanup item I still consider outstanding.

Taxuspt:
Closed: the improvements from this PR (packaging, pytest path, server startup tests) were applied and merged via PR #242. The mcp<2 cap it also contained was already merged separately via PR #227. Thank you for the contribution @vedantwalia!

PR #250, opened 2026-08-07
Add Garmin gear notes to get_gear output

victory-c:
## Summary

- Add an always-present `notes` field to every gear item returned by `get_gear`.
- Preserve all existing gear fields, statistics, sorting, and default-activity associations.
- Emit debug-only diagnostics when the optional v2 enrichment is unavailable or malformed.
- Document the new output field in both the tool docstring and README.
- Add regression coverage for populated, empty, and null/missing notes, UUID normalization, and v2 endpoint failure.

## Upstream behavior and implementation

The project's pinned `python-garminconnect==0.3.2` implementation of
[`get_gear(userProfileNumber)`](https://github.com/cyberjunky/python-garminconnect/blob/0.3.2/garminconnect/__init__.py#L2240-L2245)
uses Garmin's legacy `/gear-service/gear/filterGear` response. That response does
not contain the free-text Notes value displayed by Garmin Connect.

The Notes value is exposed as the optional lowercase `notes` field by
`/gear-service/gear/v2/list`. The v2 response uses hyphenated UUIDs while the
legacy response uses the same UUIDs without hyphens. `get_gear` now performs one
v2 request, normalizes UUIDs for the join, and adds the corresponding Notes value
to the existing curated legacy item.

The legacy response remains the canonical inventory because the v2 list can omit
retired gear. If the v2 endpoint is unavailable, an item is not present, or its
Notes value is absent/null, the item is still returned with `"notes": null`.
An explicit empty string is preserved.

## Backward compatibility

- No tool arguments or existing output keys are changed.
- The v2 call is additive and failure-tolerant.
- Existing gear inventory and optional statistics remain available if the Notes
  endpoint fails.

## Documentation

- Updated the `get_gear` tool docstring with the field and null behavior.
- Added a README section describing the `notes` field and compatibility semantics.

## Testing

- `uv run pytest tests/integration -q` — **378 passed**
- `uv run pytest tests/unit -q` — **118 passed**

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Garmin's API is accessed via the awesome [python-garminconnect](https://github.c
 - View body composition data
 - Track training status and readiness
 - Access cycling FTP and lactate threshold metrics
-- Manage gear and equipment
+- Manage gear and equipment, including free-text notes returned by `get_gear`
 - Access workouts and training plans
 - Inspect detailed workout step structures, including repeat groups and swim pace targets
 - Weekly health aggregates (steps, stress, intensity minutes)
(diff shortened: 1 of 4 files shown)


victory-c:
bro, really appreciate collaborating with you on this repo, let's keep up the great work and make more people enjoy garmin connect with agents!!!

PR #267, opened 2026-08-26
feat(training): expose Running Tolerance and heat/altitude acclimation

pajireg:
Adds two read-only training tools backed by endpoints `python-garminconnect` already supports but this server does not surface.

## `get_running_tolerance`

Wraps `metrics-service/metrics/runningtolerance/stats`.

- `aggregation="weekly"` (default) — per week: distance, impact load, and the tolerance that applied
- `aggregation="daily"` — trailing-7-day (acute) load against acute tolerance, plus Garmin's own verdict (`ABOVE_TOLERANCE`, `ABOVE_TOLERANCE_ONE_HARD_RUN`, `HIGH_LOAD`, `MEDIUM_LOAD`, `LOW_LOAD_7DAYS`)

`pct_of_tolerance` is computed in the tool rather than left to the caller, and deliberately not clamped — being over tolerance has to read as over 100.

Unlike a rolling maximum, tolerance **decays when running stops**, so it stays meaningful after a break. The docstring states plainly that this is a proprietary Firstbeat metric with no published validation, so it reads as a personalised reference rather than an established threshold.

## `get_acclimation`

Surfaces the `heatAltitudeAcclimation` section of `get_max_metrics`, which no existing tool returns: `heat_acclimation_percent`, `heat_trend`, and the altitude equivalents. The `previous_*` readings come back alongside the current ones so the direction of travel is visible in a single call, with `heat_acclimation_change` derived from the pair.

VO2 max is intentionally not duplicated here — `get_training_status` and `get_vo2max_trend` already cover it.

## Tests

7 integration tests in `tests/integration/test_running_tolerance_tools.py`, following the existing FastMCP + mocked-client pattern. Beyond the happy paths they cover:

- zero or missing tolerance not raising `ZeroDivisionError`
- a null `altitudeTrend` being dropped rather than serialised as `None`
- rows returned out of order being sorted
- an invalid `aggregation` being rejected before the API call
- the empty-result and missing-section messages

Full suite: 516 passed.

Both tools were also exercised against a live account during development.

Note: stacked on top of #258 only in that both come from the same fork; this branch touches `training.py` and a new test file, so it should merge independently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Garmin's API is accessed via the awesome [python-garminconnect](https://github.c
 - Access cycling FTP and lactate threshold metrics
 - Manage gear and equipment
 - Access workouts and training plans
+- List races and events on your Garmin Connect calendar
 - Inspect detailed workout step structures, including repeat groups and swim pace targets
 - Weekly health aggregates (steps, stress, intensity minutes)
 - Advanced cycling analytics: power zones, FIT file analysis, DI2 electronic shift intelligence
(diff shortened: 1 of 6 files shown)


Taxuspt:
Closed in favor of #289 (running tolerance with trend) and #290 (acclimation). Thank you for the contribution!

PR #268, opened 2026-08-27
fix: support Garmin Connect China authentication

cwwcn:
## Summary

Fix Garmin Connect China authentication on the current `main` branch.

The existing CN authentication flow can fall back to the retired
`mobile.integration.garmin.com` endpoint and fail with TLS EOF errors, leaving
the `JWT_WEB` cookie unset after ticket consumption.

This PR brings over the CN compatibility work from #249 while keeping the
change focused on authentication and dependency updates. Documentation
rebranding, DXT metadata/binary changes, and other unrelated files from #249
are intentionally excluded.

## Validation

- `uv run pytest tests/unit tests/integration`
- 501 tests passed

## Credit

Implementation is based on the work in #249. This PR is intended as a smaller,
focused version that is easier to review and merge.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,8 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
     "python-dotenv==1.2.2",
-    "garminconnect==0.3.2",
+    "garminconnect==0.3.2; python_version < '3.12'",
+    "garminconnect==0.3.4; python_version >= '3.12'",
     "requests==2.33.0",
     # mcp 2.x renamed mcp.server.fastmcp -> mcp.server.mcpserver (FastMCP ->
     # MCPServer). Cap until the server is ported to the new API.
@@ -32,4 +33,3 @@ dev-dependencies = [
     "pytest-mock>=3.14.0",
     "pytest-timeout>=2.3.1",
 ]
-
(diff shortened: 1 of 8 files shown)


mknightz:
Comparative review of the three China-auth PRs (#180, #249, this one), for merge-triage:

**#180 is superseded and should be closed rather than merged.** It monkey-patches `garminconnect==0.3.2`; upstream 0.3.4 (DI OAuth) and 0.3.5 (mobile service URLs) made the client domain-aware natively, and #214 bumps the pin to 0.3.7 where the patched `.com` constants have no call sites. pn's analysis in the #180 thread reaches the same conclusion — its own tests would also fail post-bump because the patch version-gates itself to 0.3.2.

**Between #249 and #268, #268 is the right merge candidate:**

- same `garmin_cn_compat.py` core (this PR is explicitly derived from #249 and credits it), but **focused on auth + dependency files only** — #249 additionally carries README rebranding, DXT manifest/binary changes, and other unrelated churn that widens the review surface;
- #268's suite passes 501 tests on the current base per its validation notes, vs. #249's mixed Python-version matrix (which drops CN support to 3.12-only on 3.10/3.11 installs — a policy decision the focused PR defers);
- after #214 lands (0.3.7 domain-aware URLs), the remaining *repo-side* value of both PRs is the same narrow piece: **persisting/restoring `JWT_WEB` + CSRF cookie state** when DI exchange is unavailable — upstream's token serializer keeps only DI tokens, so a fallback session is lost across restarts. #268 carries exactly that.

**Suggested sequence:** merge #214 (bump, verified against current main — see my review there), then #268 for the cookie-persistence gap, then close #180 and #249 as superseded. If the maintainer prefers one PR covering everything, #249 is the complete version — but the narrower diff of #268 is the faster, safer merge.

PR #269, opened 2026-08-28
fix(server): start Garmin login in the background instead of blocking the MCP handshake

### LINKED ISSUE #255: Initialize times out after 60s — Garmin login appears to block the handshake (macOS, Claude Desktop + MCP Inspector)
**Environment**
- macOS (Apple Silicon), Python 3.12
- Server launched via `uvx --from git+https://github.com/Taxuspt/garmin_mcp garmin-mcp`, main @ 3610be6
- Clients tested: Claude Desktop and MCP Inspector — both fail identically

**What happens**

The client sends `initialize`. The server prints `Trying to login to Garmin Connect using token data from directory ...` and then never answers. Exactly 60 seconds later the client sends `notifications/cancelled` and drops the connection. Claude Desktop shows "Could not attach to MCP server garmin" and no tools are registered.

The line `Garmin Connect client initialized successfully.` never appears under either client — but it always appears when the server is started directly in a terminal.

**Log**

```
08:21:23.332  Server started and connected successfully
08:21:23.806  Message from client: method="initialize" id=0
              Trying to login to Garmin Connect using token data from directory '/Users/<user>/.garminconnect'...
08:22:23.801  Message from client: method="notifications/cancelled"
08:22:23.802  Client transport closed
```

MCP Inspector console export for the same run contains a single line:

```json
[{ "timestamp": "2026-08-13T08:27:56.819Z",
   "message": "Trying to login to Garmin Connect using token data from directory '/Users/<user>/.garminconnect'..." }]
```

**Ruled out**

- Tokens: `garmin-mcp-auth --verify` reports valid tokens
- Stripped environment: `env -i HOME=/Users/<user> ... garmin-mcp` starts fine and logs in successfully
- stdin: `garmin-mcp < /dev/null` logs in successfully, so the login path does not read from stdin
- stdout pollution: with `> /dev/null` the status lines are still visible, so they go to stderr, not stdout
- Stale local checkout: same behaviour from a local clone via `uv run` and from `uvx` against main
- Duplicate processes: `ps aux | grep garmin` returns nothing before the failing start

So: the server works standalone under every condition I could reproduce from the client side, and fails under any MCP client.

**Hypothesis**

The Garmin login appears to run synchronously before the server replies to `initialize`, so the handshake stays unanswered while the login is in flight. In a terminal that is invisible because nothing is waiting on a reply; under an MCP client it exceeds the 60-second initialize timeout and the connection is torn down. Deferring the login until after the initialize response (or moving it into a lazy/first-tool-call path) would likely fix it.

Happy to run further tests or provide full logs. #

Note: this is not the Windows CRLF/stdio issue discussed in #128 — there is no JSON parse error here, the client simply never receives a response to initialize and times out after 60s. macOS does not do newline translation on stdout.

Oneengineer87:
## Summary

Closes #255.

`main()` currently calls `init_api()` (a synchronous, network-bound Garmin
login) before `app.run(transport=...)`. Under stdio transport, the process
doesn't start reading/responding to MCP protocol messages until `app.run()`
begins, so a client's `initialize` request sits unread for the full duration
of login. Most MCP clients (Claude Desktop, MCP Inspector) give up after 60s
and tear down the connection — even though `init_api()` usually finishes in
well under a second (verified: ~0.8s against a cached token in this repro).

This PR moves Garmin login to a background thread so `app.run()` — and the
MCP handshake — starts immediately, regardless of login time. Tool calls
transparently wait on login (via a small `_PendingGarminClient` stand-in
passed to the existing `_GarminProxy`) instead of the handshake waiting on
it, and fail with one clear, actionable error if login doesn't finish within
90s or fails outright.

Since `sys.stdout` is a single process-wide object, a `_ThreadFilteredStream`
is installed as `sys.stdout` before the login thread starts, so any stray
output from that thread (e.g. `garminconnect` progress dots) can't land in
the middle of the JSON-RPC messages the main thread writes — this also let
me remove two now-redundant local `sys.stdout` swaps inside `init_api()`
that predate this change and would otherwise race with it.

Scope: stdio transport only (matches the bug report/repro); no change to
`streamable-http` / `sse` startup or to how login itself works (token
loading, credential fallback, MFA) — only *when* it happens relative to
server startup.

## Test plan

- [x] New unit tests for `_ThreadFilteredStream` and `_PendingGarminClient`
      (owner-thread-only writes, blocks-then-delegates, failure/timeout
      error messages, composes with `_GarminProxy` unchanged)
- [x] New regression test (`test_main_starts_server_before_garmin_login_completes`)
      that fails against the old code (login must not need to finish before
      `app.run()` is called) and passes after the fix
- [x] Full existing suite passes unmodified (503 passed)
- [x] Manually confirmed `init_api()` itself is unaffected and still logs in
      correctly (~0.8s, cached token)

### DIFF
--- a/src/garmin_mcp/__init__.py
+++ b/src/garmin_mcp/__init__.py
@@ -186,6 +186,68 @@ class _ThreadFilteredStream:
         return getattr(self._real_stream, name)
 
 
+class _PendingGarminClient:
+    """Stands in for the real Garmin client until a background login finishes.
+
+    Passed as the ``client`` argument to ``_GarminProxy``, so ``_GarminProxy``
+    itself needs no changes: every attribute it looks up on ``self._client``
+    comes through here first. Before login finishes, that lookup blocks (up
+    to ``timeout`` seconds, ``0``/``None`` disables the bound) instead of
+    returning immediately -- this is what lets ``main()`` call ``app.run()``
+    right away instead of waiting on Garmin login before the MCP handshake
+    can be answered (issue #255).
+    """
+
+    def __init__(self, timeout):
+        self._timeout = timeout
+        self._ready = threading.Event()
+        self._client = None
+        self._login_error = None
+
+    def start(self, login_fn):
+        """Run ``login_fn`` on a background daemon thread; store its outcome."""
+
+        def _worker():
+            try:
+                client = login_fn()
+            except BaseException as exc:  # noqa: BLE001 - stored, not raised here
+                self._login_error = exc
+            else:
+                if client is None:
+                    self._login_error = RuntimeError(
+                        "Garmin login failed. Run 'garmin-mcp-auth' to "
+                        "authenticate, then restart the server."
+                    )
+                    print(
+                        "Garmin Connect client failed to initialize (see "
+                        "errors above). Tool calls will fail until this is "
+                        "fixed; run 'garmin-mcp-auth' and restart the server.",
+                        file=sys.stderr,
+                    )
+                else:
+                    self._client = client
+                    print(
+                        "Garmin Connect client initialized successfully.",
+                        file=sys.stderr,
+                    )
+            finally:
+                self._ready.set()
+
+        threading.Thread(target=_worker, name="garmin-login", daemon=True).start()
+        return self
+
+    def __getattr__(self, name):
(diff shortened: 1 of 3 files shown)


Taxuspt:
Applied and merged via #280. Thank you!

Issue #264, opened 2026-08-25
get_training_status fails with 'NoneType' object has no attribute 'get'


skweeker:
get_training_status fails for my Garmin account even though Garmin Connect currently has valid Training Status data for the account.

**Error**

Error retrieving training status data: 'NoneType' object has no attribute 'get'

**Reproduction**

Call:

get_training_status(date="2026-08-24")

**Expected behavior**

The tool should return the current Training Status information. Garmin Connect currently displays Training Status data for this account, including metrics such as:

* Training Status
* Acute / 7-day training load
* Optimal load range
* Load focus

Other MCP endpoints

Other Garmin MCP functionality is working correctly for the same account. I’ve successfully retrieved data using endpoints including:

* get_sleep_summary
* get_activities
* get_activity
* get_training_effect

Individual activities also successfully return Garmin-specific metrics including aerobic/anaerobic Training Effect and activity training load.

**Environment**

I’m accessing garmin_mcp through the hosted MissingMCP Garmin connector.

MissingMCP indicates that its Garmin service runs the upstream Taxuspt/garmin_mcp server unmodified, so I’m reporting the parsing failure here.

**Possible cause**

Based on the exception, it looks like one of the nested values returned by Garmin may be None/null or absent, while the curated Training Status parser assumes it is an object and calls .get() on it.

It would be useful if missing Garmin fields could be handled gracefully rather than causing the entire get_training_status call to fail.

If useful for debugging, I’m happy to provide additional information about the Garmin devices/account or test additional calls.

PR #271, opened 2026-08-29
fix(training): handle None values in get_training_status device data

armand33:
Fix issue #264 where NoneType object has no attribute get was raised when Garmin API returned None for device data.

The fix ensures device_data and load_data are always dicts by using data or {} pattern when iterating over device entries.

Add regression tests for both null device data and null load balance data.


armand33:
Closed as duplicate - issue #264 is already addressed by PR #265

PR #274, opened 2026-08-30
fix(workouts): emit repeat groups so multi-set strength exercises keep their set count (closes #273)

### LINKED ISSUE #273: create_strength_workout ignores 'sets' — every exercise uploads as 1 set
## Symptom

Creating a strength workout with multi-set exercises, e.g.

```json
{"name": "Goblet Squat", "sets": 4, "reps": 8, "rest_seconds": 90}
```

uploads fine, but Garmin Connect (and the watch) shows every exercise as **1 set** (1/1). The requested set count only survives in the step description text.

## Cause

`build_strength_json` reads `sets` and then uses it exclusively in the description string:

https://github.com/Taxuspt/garmin_mcp/blob/3610be6feed93088d85b0f35aba9d7d07c2505a7/src/garmin_mcp/workout_builders.py#L281

Each exercise is emitted as a single reps-based `ExecutableStepDTO`; there is no `RepeatGroupDTO`, so Garmin has nothing to iterate.

## Expected

`sets: N` (N > 1) should produce a repeat group with `numberOfIterations: N` containing the work step and the rest step, the same structure `build_walk_run_json` already uses for intervals — and the same `stepTypeId 6` / `endCondition iterations` shape `_sanitize_repeat_group` enforces on the upload path.

PR with fix + tests incoming.

cmatte:
Closes #273.

## What

`build_strength_json` read each exercise's `sets` and used it only in the description text, so every strength exercise uploaded as a single reps step and Garmin displayed **1 set** regardless of the request.

An exercise with `sets > 1` is now emitted as a `RepeatGroupDTO` iterated `sets` times, with the work step and rest step inside the group (rest after every set — and therefore no extra inter-exercise rest step for grouped exercises). Single-set exercises keep the existing flat shape byte-for-byte, including the inter-exercise rest behaviour.

The group carries the explicit `stepTypeId 6` / `endCondition: iterations` shape that `_sanitize_repeat_group` already enforces on the upload path, since the Garmin API silently corrupts repeat groups lacking a valid iterations endCondition.

## Before / after (one exercise, 4×8, 90 s rest)

Before: one `ExecutableStepDTO` (reps 8) — Connect shows 1 set.
After:

```json
{
  "type": "RepeatGroupDTO",
  "stepType": {"stepTypeId": 6, "stepTypeKey": "repeat"},
  "numberOfIterations": 4,
  "smartRepeat": false,
  "endCondition": {"conditionTypeId": 7, "conditionTypeKey": "iterations"},
  "workoutSteps": [
    {"type": "ExecutableStepDTO", "stepType": {"stepTypeKey": "interval"}, "endCondition": {"conditionTypeKey": "reps"}, "endConditionValue": 8.0},
    {"type": "ExecutableStepDTO", "stepType": {"stepTypeKey": "recovery"}, "endCondition": {"conditionTypeKey": "time"}, "endConditionValue": 90.0}
  ]
}
```

## Tests

- `test_build_strength_json_structure` updated for the grouped shape
- new: single-set stays flat; multi-set group shape (iterations endCondition, rest inside); `rest_seconds: 0` group has no recovery step
- `_work_steps` helper flattens repeat groups so the category tests keep exercising work steps
- `tests/unit`: 121 passed

### DIFF
--- a/src/garmin_mcp/workout_builders.py
+++ b/src/garmin_mcp/workout_builders.py
@@ -261,11 +261,14 @@ def build_strength_json(
 ) -> dict:
     """Build the Garmin Connect JSON for a strength workout.
 
-    Each exercise becomes a reps-based work step. The name is preserved in the step
-    "description", which is what survives the round trip. It is also sent as
-    "exerciseName", but Garmin only retains that when it matches one of its own
-    exercise keys (e.g. "FARMERS_CARRY"); any other value is accepted and then
-    stored as an empty string.
+    Each exercise becomes a reps-based work step. When "sets" > 1 the exercise is
+    emitted as a RepeatGroupDTO iterated that many times (work + rest per set), so
+    Garmin shows the real set count — previously "sets" only ever reached the
+    description text and every exercise displayed as 1 set. The name is preserved
+    in the step "description", which is what survives the round trip. It is also
+    sent as "exerciseName", but Garmin only retains that when it matches one of
+    its own exercise keys (e.g. "FARMERS_CARRY"); any other value is accepted and
+    then stored as an empty string.
 
     "category" is optional and only emitted when the caller supplies one, uppercased
     and otherwise passed through untouched. Garmin validates it against its own enum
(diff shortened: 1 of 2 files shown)


Taxuspt:
Applied and merged via #279. Thank you!

PR #275, opened 2026-09-01
fix(training): carry VO2 max values forward so trend matches the Connect chart

### LINKED ISSUE #261: get_vo2max_trend
get_vo2max_trend returns incomplete historical data.
For the period March–August 2026, Garmin Connect’s web UI shows monthly VO₂ Max values and a clear progression: approximately 33.8–34.1 in March, 34.5 in April, 35.0 in May–June, followed by a gradual decline to 34.6 in August.
However, get_vo2max_trend returned only sparse fallback points:
- 1 March: 34.0 (get_training_status)
- 30 April: 35.0 (get_max_metrics)
- 23 May: 35.0 (get_training_status)
The endpoint therefore misses most available historical values and can incorrectly report no change during periods where Garmin Connect displays a trend.
A workaround is to call get_training_status for each relevant date. That endpoint returns vo2_max_precise and matched the Garmin chart much more closely—for example, 34.5 on 30 April, 35.3 in mid-July, and 34.6 in mid-August.
Expected behavior: get_vo2max_trend should retrieve the complete historical VO₂ Max series exposed by Garmin Connect, rather than combining a small number of values from summary/fallback endpoints.

mknightz:
## Summary

- `get_vo2max_trend` returned only sparse points (reported in #261) because Garmin's max-metrics range endpoint (`metrics-service/metrics/maxmet/daily/{start}/{end}`) records an entry **only on days with a VO2 max recompute** (i.e. after an activity), while the Connect web chart carries each value forward until the next one.
- Add `_build_vo2_trend_series`: expands the sparse measured days into a dense daily series from the first measurement through the requested end date, marking carried days with `"carried_forward": true`. This matches the chart the reporter compares against.
- `_extract_vo2_measurements` now prefers `vo2MaxPreciseValue` over `vo2MaxValue`: the latter is rounded to 0.5 while the chart and per-date `get_training_status` (which the reporter verified matches the chart) report the 0.1-precision estimate.
- Document the carry-forward semantics in the tool docstring.

## Why

Closes #261. Per-date max-metrics data exists only for recompute days (see cyberjunky/python-garminconnect#253: "until you've done an activity on a given day, there are no Max Metrics"), so the old loop's per-day sources could only ever produce the few sparse fallback points the reporter saw. Carrying values forward reconstructs the full series from the same single range request.

The per-day `get_training_status` fallback is unchanged: it is still consulted for days the range response does not cover (the reporter confirmed it returns accurate per-date values), and forward-fill only applies when neither source has data for a day.

## Test plan

- `uv run pytest tests/unit tests/integration`: 498 passed (includes 3 new unit tests for the series builder + precise-value preference and 1 new integration test covering the sparse-recompute-days scenario end to end).
- Full `uv run pytest`: all remaining failures are pre-existing live-credential e2e tests (`OAuth tokens not found`), unrelated to this change.


Taxuspt:
Applied and merged via #283. Thank you!

PR #292, opened 2026-09-02
fix(workouts): build real repeat groups for strength sets

edwardlg96:
## Problem

`build_strength_json` reads the `sets` value from each exercise but only interpolates it into the step description. No `RepeatGroupDTO` is ever emitted, so a `{"sets": 3, "reps": 12}` exercise uploads as a **single 12-rep step** whose description reads `"Back Squat: 3 sets x 12 reps"`.

The watch counts one set while the description promises three. Confirmed by reading an uploaded workout back:

```json
{
  "order": 1,
  "type": "interval",
  "description": "Sentadilla peso corporal: 3 sets x 12 reps",
  "end_condition": "reps",
  "end_condition_value": 12.0
}
```

## Fix

Each exercise with more than one set now becomes a `RepeatGroupDTO` of `sets` x (work step + rest step), so the set count is structural and the watch counts it down. Single-set exercises keep the previous flat shape, with the rest as a sibling.

Two supporting changes:

- **Step ordering.** `stepOrder` is now assigned by a recursive walk, so nested steps continue the outer numbering rather than restarting — which is what Garmin expects.
- **Adjacent bug.** The "skip the rest after the last exercise" check compared exercise dicts by value (`ex != exercises[-1]`), so two identical exercises made the earlier one lose its rest step. It now compares by index.

The `sets` label is dropped from the step description, since the count is no longer text.

## Verification

Uploaded to a real account and read back — `repeat_count: 3` on every group, with the reps step and rest step nested inside.

Four existing tests encoded the old flat-step behaviour and were updated. Four new tests cover the repeat group and its `endCondition`, the single-set shape, sequential step ordering through groups, and the identical-exercises case. Full suite: **498 passed**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,10 +138,12 @@ Returns: `{"status": "success", "workout_id": 1234567890, ...}`
 
 ### `create_strength_workout`
 
-Creates a strength workout from a list of exercises. Each becomes a reps-based step, with the
-name kept in the step description. The name is also sent as `exerciseName`, but Garmin only
-retains that when it matches one of its own exercise keys (e.g. `FARMERS_CARRY`) — any other
-value is accepted and then stored empty.
+Creates a strength workout from a list of exercises. An exercise with more than one set
+becomes a repeat group of `sets` x (reps step + rest step), so the watch counts the sets
+down; a single-set exercise stays a flat step with its rest alongside. The name is kept in
+the step description. The name is also sent as `exerciseName`, but Garmin only retains that
+when it matches one of its own exercise keys (e.g. `FARMERS_CARRY`) — any other value is
+accepted and then stored empty.
 
 `category` is optional and passed straight through. Omit it and the key is left out of the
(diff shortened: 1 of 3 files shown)


Taxuspt:
Thanks for this! The core issue here — no RepeatGroupDTO being emitted for multi-set strength exercises — was already fixed on main in b8ea726 (fix(workouts): emit repeat groups so multi-set strength exercises keep their set count), merged the day before this PR was opened, so your fork was likely based on a slightly older commit. Re-applying this diff on top of current main produces conflicting/duplicated step-building logic rather than a clean merge.

One thing this PR still catches that's worth a follow-up: the rest-step skip check on the single-set path compares `ex != exercises[-1]` by value rather than by index, so two exercises with identical dicts could incorrectly drop the rest step on a non-final exercise. I'll track that separately. Thanks for digging into this!

PR #300, opened 2026-09-06
fix: surface the original Garmin error and bump garminconnect to 0.3.11

### LINKED ISSUE #299: "Garmin Connect is unreachable" wrapper message hides real error; garminconnect pinned to 0.3.2
Two related things, both hit today.

**1. Misleading error mapping**

`src/garmin_mcp/__init__.py` maps every `GarminConnectConnectionError` to:

```
Garmin Connect is unreachable. Check your network connection or try again later.
```

But the library raises that class for non-network problems too. Today `get_rhr_day`, `get_race_predictions` and every other display-name-dependent tool failed with "unreachable" while `get_activities` / `get_user_profile` worked fine in the same session. The real cause was `display_name` being `None` after token login (cyberjunky/python-garminconnect#421). Restarting the server fixed it.

Suggestion: keep the friendly prefix but don't replace the original message — the `(Details: ...)` suffix is what made it debuggable, so promote it rather than burying it behind a network-error claim.

**2. Dependency pin**

`pyproject.toml` pins `garminconnect==0.3.2`; upstream is at 0.3.11 and has already fixed at least one variant of the above (`resume_login` no longer swallows the profile-fetch failure). Worth bumping.

Great project otherwise — 110+ tools and they just work.

MishaKav:
Closes #299

`_GarminProxy` replaced every `GarminConnectConnectionError` with "Garmin Connect is unreachable…" and tucked the real cause into a "(Details: …)" suffix, so failures unrelated to the network (for example a missing display name) were reported as connectivity problems. The project was also pinned to garminconnect 0.3.2 while 0.3.11 is current.

Commit 1 turns each `_MESSAGES` entry into a (prefix, hint) pair and builds the message as `<prefix>: <original>. <hint>`, so the library's own text always leads and the hint follows. Commit 2 bumps garminconnect to 0.3.11; every release after 0.3.2 requires Python 3.12, so `requires-python` and the CI matrix move to 3.12/3.13 (the Dockerfile already uses 3.12). 0.3.11 adds `GarminConnectNotFoundError` as a subclass of the connection error, so it gets its own mapping to avoid reporting a 404 as "unreachable". `uv.lock` was regenerated with `uv lock`; unit and integration suites pass (589). Commit 2 is self-contained and can be dropped if you would rather keep the 3.10 floor. Not touched: `dxt/manifest.json` still says `python >=3.10`; changing it requires rebuilding `garmin-mcp.dxt`, which I left to the release process.


Taxuspt:
Cherry-picked just the error-message fix (first commit) as #314, which is merged now — that resolves #299 without pulling in the garminconnect 0.3.11 / Python 3.12 bump bundled in this PR's second commit. That version bump touches .github/workflows/ci.yml (can't be merged via the API) and is also a standing open design question here (garminconnect 0.3.7/0.3.8 upgrade + Python 3.12 minimum are both already flagged) — leaving this PR open for the maintainer to decide on that part. Thank you!

victory-c:
Noting this is still conflicting with main. Opened #345 to get a maintainer decision on the garminconnect baseline, and #341 for broader overlapping-PR triage.

PR #324, opened 2026-09-12
fix(training): fall back to activity list when training effect details 403

### LINKED ISSUE #304: get_training_effect returns HTTP 403 for valid, recent activity IDs
### Description
`get_training_effect(activity_id=...)` consistently fails with a 403 Forbidden error, even when called with valid, recent activity IDs that exist and are accessible via other tools.

### Steps to reproduce
Called `get_training_effect` with three different activity IDs, all recent (within the last week) and all confirmed valid (visible via other tools like activity listing):

1. Activity ID from a long run
2. Activity ID from a threshold session
3. Activity ID from an easy run

All three calls returned the same error:

Error retrieving training effect data: Garmin Connect is unreachable. Check your network connection or try again later. (Details: HTTP error: API Error 403 - HTTP 403 Forbidden)

### Expected behavior
Training effect data returned for each activity.

### Actual behavior
403 Forbidden on every call, regardless of activity type or age.

### Other context
- Other tools in the same session worked fine in parallel (e.g. `get_training_status`), so this doesn't appear to be a broader auth/token issue, it seems isolated to this specific endpoint.
- Setup: Forerunner 255, server installed via `uvx --from git+https://github.com/Taxuspt/garmin_mcp`.
- The 403 message text ("Garmin Connect is unreachable... check your network connection") seems misleading for what's actually a permissions/auth-style error code, worth checking if that wrapping is masking the real cause.

victory-c:
## Summary
- `get_training_effect` currently calls `/activity-service/activity/{id}`, which returns HTTP 403 for some valid, recently synced activities even while list tools work.
- On 403, look the same activity up in activity list search and curate aerobic/anaerobic training effect, load, and label from that payload.
- If the activity is not in the recent-search window, return a clear 403 message instead of calling it a network failure.

Closes #304

## Test plan
- [x] `uv run pytest tests/unit/test_training_effect.py tests/integration/test_training_tools.py`
- [ ] With a live Garmin account that previously 403'd: `get_training_effect` on a recent activity returns TE fields with `"source": "activity_list"`
- [ ] A normal activity that still allows details keeps `"source": "activity_details"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -882,6 +882,17 @@ Once connected in Claude, you can ask questions like:
 
 ## Troubleshooting
 
+### `get_training_effect` returns HTTP 403 for a valid activity
+
+Garmin's activity-details endpoint (`/activity-service/activity/{id}`) can
+return **403 Forbidden** even when the same activity is visible in list tools.
+`get_training_effect` now falls back to activity list search, which still
+includes aerobic/anaerobic training effect for recent activities.
+
+If the activity is older than the recent-search window, list the activity with
+`get_activities` / `get_activities_by_date` and retry, or use those list fields
+directly.
+
 ### "Failed to spawn process: No such file or directory"
 
 If Claude Desktop can't find `uvx`, it's because `uvx` is not in the PATH that Claude Desktop uses. To fix this:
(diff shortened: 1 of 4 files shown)


victory-c:

PR #325, opened 2026-09-12
fix(challenges): return Connect UI goals from get_goals

### LINKED ISSUE #317: get_goals returns no goals when active goal exists in Garmin Connect
`get_goals` does not return goals that are visible in Garmin Connect.

I currently have an active cycling goal in Garmin Connect:

- **Goal name:** GCC 2026
- **Activity type:** Cycling
- **Target:** 100 miles
- **Start date:** 2026-09-01
- **Status:** Active

Garmin Connect currently shows progress toward the goal, so the goal is active and being updated normally.

## Steps to reproduce

Call:

get_goals(goal_type="active")

## Actual result

The tool returns:

No active goals found.

I also tested:

get_goals(goal_type="future")
get_goals(goal_type="past")

These return:

No future goals found.
No past goals found.

So this does not appear to be an issue where the active goal is simply being classified under the wrong goal type. No goals are being returned at all.

## Expected result

`get_goals(goal_type="active")` should return the active "GCC 2026" cycling goal and its current progress.

## Additional information

At the time of testing, Garmin Connect web showed:

- Target: 100.00 mi
- Remaining: 91.29 mi
- Complete: 8%
- Start date: 09/01/2026
- Time left: 22 days

A screenshot from Garmin Connect showing the active goal can be provided/attached.

Image

victory-c:
## Summary
Garmin Connect's current Goals UI (named distance/time targets, e.g. a monthly 100-mile cycling goal) is backed by `/goal-service/goal/goals?status=…`, the same endpoint python-garminconnect calls. Two things decide whether those goals come back, per the live testing in cyberjunky/python-garminconnect#431:

1. **`Sec-Fetch-Site: same-origin`** must be sent. Without it the list is silently `[]`. garminconnect 0.3.16 adds this.
2. **`start` is 1-based.** `start=0` returns `[]` even when goals exist. garminconnect 0.3.16 still starts at 0, so `get_goals()` keeps missing them.

`get_goals` now calls the endpoint directly the way Connect's Goals page does: `status`, `start=1`, `limit=100`, `sortOrder=asc`, and the header. It paginates by `limit` and stops on a short page or a repeated page. Connect UI goals are flattened into target/progress/remaining fields. Older wellness goals pass through unchanged. The library `get_goals()` is still used as a fallback if the direct call errors or returns nothing. This works on the currently pinned `garminconnect==0.3.2`, so no dependency bump is needed.

This replaces the earlier socialProfile/`userId` probing. Garmin doesn't need a `userId` here.

Closes #317

## Test plan
- [x] `uv run pytest tests/integration tests/unit` (638 passed)
- [x] Unit tests model the observed server contract: the header is required, `start=0` returns `[]`, 1-based pagination works, the loop stops if `start` is ignored, and the library fallback is used.
- [x] Live run against an account with no goals: all three statuses return `[]` with no 400s, and the tool reports "No … goals found."
- [ ] Live account with a Connect UI goal (@skweeker): `get_goals(goal_type="active")` returns the named goal with its progress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -882,6 +882,14 @@ Once connected in Claude, you can ask questions like:
 
 ## Troubleshooting
 
+### `get_goals` returns no goals that exist in Garmin Connect
+
+Garmin Connect's current Goals UI (named distance/time targets such as a
+100-mile cycling goal) is served by `/goal-service/goal/goals` **without**
+the older `status=active` query used by python-garminconnect. `get_goals`
+now reads that unfiltered endpoint first and falls back to the legacy
+wellness-goals API when it is empty.
+
 ### "Failed to spawn process: No such file or directory"
 
 If Claude Desktop can't find `uvx`, it's because `uvx` is not in the PATH that Claude Desktop uses. To fix this:
--- a/src/garmin_mcp/challenges.py
+++ b/src/garmin_mcp/challenges.py
@@ -224,6 +224,142 @@ def _format_pr_value(value: float, value_type: str) -> str:
     return str(value)
 
 
+def _parse_goal_date(value: Any) -> Optional[datetime.date]:
+    """Parse a Garmin goal date string (YYYY-MM-DD) into a date, or None."""
+    if not isinstance(value, str) or len(value) < 10:
+        return None
+    try:
+        return datetime.datetime.strptime(value[:10], "%Y-%m-%d").date()
+    except ValueError:
+        return None
+
+
+def _as_goal_list(payload: Any) -> List[Dict[str, Any]]:
+    """Normalize Connect goal payloads to a list of dicts."""
+    if isinstance(payload, list):
+        return [item for item in payload if isinstance(item, dict)]
+    if isinstance(payload, dict):
+        for key in ("goals", "items", "data"):
+            nested = payload.get(key)
+            if isinstance(nested, list):
+                return [item for item in nested if isinstance(item, dict)]
+        if any(key in payload for key in ("name", "goalType", "id", "type")):
+            return [payload]
+    return []
+
+
+def _is_connect_ui_goal(goal: Dict[str, Any]) -> bool:
+    """True for the current Connect Goals UI shape (distance/time accumulation)."""
+    progress = goal.get("progress")
+    return "name" in goal and (
+        isinstance(progress, dict)
+        or "distanceInMeters" in goal
+        or ("type" in goal and "goalType" not in goal)
+    )
+
+
+def _classify_goal(goal: Dict[str, Any], today: datetime.date) -> str:
+    """Classify a Connect UI goal as active, future, or past."""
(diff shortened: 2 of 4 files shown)


skweeker:
I tested the approach in this PR against a live account with an active Connect UI goal. Using `garminconnect==0.3.2`, calling `/goal-service/goal/goals` without `status` returns HTTP 400: `userId and status cant be null`. Adding `start`/`limit` does not change that behavior. Calls using `status=active`, `past`, or `future` all return `[]`, even when `userId` is explicitly supplied. So` _fetch_connect_ui_goals()` as currently implemented would catch the 400s and fall back to the legacy API, which still returns no goals for this account.

victory-c:
Thanks for the live check — that 400 is the actual contract now, so the previous unfiltered `/goal-service/goal/goals` call was always going to miss Connect UI goals.

`get_goals` no longer calls that endpoint without both query params. It:

1. Resolves `userId` from `/userprofile-service/socialProfile` (`id` / `profileId`, then `displayName`). python-garminconnect's `get_user_profile()` hits user-settings, not social profile, so an id from there still comes back `[]`.
2. Queries `/goal-service/goal/goals` with **both** `userId` and `status` (`active` first, then `future`/`past` if that list is empty). `userProfilePk` is tried if `userId` is empty.
3. If those are still empty, tries `/goal-service/goal/user/effective/{today}` and `/goal-service/goal/{userId}/goals`.
4. Falls back to the legacy `get_goals(status)` API only after that.

Pushed in 5d06052. I have not re-run this against a live Garmin account; if `socialProfile.id` + `status=active` is still `[]` on yours, a HAR from Connect’s Goals page would tell us the real path/params.

victory-c:

skweeker:
There's a new version of `python-garminconnect`, `0.3.16` but it does not completely solve this issue, yet.

victory-c:
Thanks @skweeker. Your findings in cyberjunky/python-garminconnect#431 (the `Sec-Fetch-Site: same-origin` header, plus `start` being 1-based) were the missing pieces. I've reworked this PR around them in b2a22a6.

`get_goals` now requests `/goal-service/goal/goals?status=<status>&start=1&limit=100&sortOrder=asc` with `Sec-Fetch-Site: same-origin`, the same way Connect's Goals page does. It no longer calls socialProfile or sends `userId`. The library `get_goals()` is only a fallback now. Because it doesn't depend on the library's pagination, it works on both the pinned 0.3.2 and 0.3.16.

My test account has no goals, so the positive case isn't verified live yet. If you have a minute to run `get_goals` from this branch against your GCC 2026 goal, that would confirm it end to end.

skweeker:
I've updated my MCP with this change and it's still not working for me. Would it be possible for you to test this with a live account @victory-c? Creating a goal in Garmin Connect is easy and it can be deleted afterwards.

PR #326, opened 2026-09-12
feat(health): expose daily recovery time remaining

### LINKED ISSUE #305: Feature request: expose recovery time remaining as a standalone daily value
### Description
Garmin devices show a "recovery time remaining" figure (hours until fully recovered from recent training load, based on Firstbeat/EPOC), visible directly on-device and in Garmin Connect. There's currently no tool in this server that exposes it as a standalone daily value.

This is distinct from Training Readiness (which isn't available on all devices, e.g. Forerunner 255) and distinct from Training Status/ACWR (which is a longer-term load ratio, not the same acute recovery clock). Recovery time updates per-activity and decays over the following 24 to 48 hours.

### What I've tried
- `get_training_status` returns acute/chronic load and ACWR, but not recovery hours.
- `get_training_readiness` and `get_morning_training_readiness` return empty for devices without the Training Readiness feature.
- `get_training_effect` is scoped to a specific activity ID rather than "today's current recovery time," and is also currently broken (see separate issue).

### Request
A tool (something like `get_recovery_time` or similar) that returns the current recovery time remaining in hours, independent of whether the device supports the full Training Readiness score. This would be useful for anyone building a custom readiness composite, since recovery time is one of the six standard inputs Garmin uses for its own Training Readiness formula and is a meaningful signal on its own even without the rest of that feature set.

### Context
Setup: Forerunner 255, server installed via `uvx --from git+https://github.com/Taxuspt/garmin_mcp`.

victory-c:
## Summary
- Garmin devices show recovery time remaining as a standalone daily clock. That value was only nested inside Training Readiness, which some devices (e.g. Forerunner 255) never publish.
- Add `get_recovery_time_remaining(date?)` returning `{remaining_hours, recovery_score, state, source}`.
- Prefer the latest Training Readiness snapshot (including `REACHED_ZERO`), then morning readiness, then decay `recoveryTime` from recent activities.

Closes #305

## Test plan
- [x] `uv run pytest tests/unit/test_recovery_time.py tests/integration/test_health_wellness_tools.py`
- [ ] On a device with Training Readiness: remaining hours match Connect.
- [ ] On a Forerunner 255 (or similar): tool either decays activity recoveryTime or returns `state: unavailable` with a clear message instead of an empty readiness payload.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Garmin's API is accessed via the awesome [python-garminconnect](https://github.c
 This MCP server implements **110+ tools** covering ~90% of the [python-garminconnect](https://github.com/cyberjunky/python-garminconnect) library (v0.3.2):
 
 - ✅ Activity Management (20 tools) - includes write tools for type, description, event type, perceived effort, and feel
-- ✅ Health & Wellness (33 tools) - includes custom lightweight summary tools
+- ✅ Health & Wellness (34 tools) - includes custom lightweight summary tools
 - ✅ Training & Performance (13 tools) - includes CTL/ATL/TSB, HRV, VO2 max, and respiration trends
 - ✅ Workouts (8 tools)
 - ✅ Devices (7 tools)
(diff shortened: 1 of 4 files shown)


victory-c:

PR #327, opened 2026-09-12
feat(activity): add delete_activity with a confirmation gate

victory-c:
## Summary
- `delete_activity()` was intentionally skipped as a destructive operation.
- Add the tool with a confirmation gate: the default call returns a preview (`status: needs_confirmation`); a second call with `confirm=true` performs the delete.
- Log each deletion to stderr for an audit trail. There is no undo.

## Test plan
- [x] `uv run pytest tests/unit/test_delete_activity.py tests/integration/test_activity_management_tools.py`
- [ ] Live: call `delete_activity` without `confirm` and verify nothing is deleted
- [ ] Live: call again with `confirm=true` on a disposable test activity and confirm it disappears from Connect

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Garmin's API is accessed via the awesome [python-garminconnect](https://github.c
 
 This MCP server implements **110+ tools** covering ~90% of the [python-garminconnect](https://github.com/cyberjunky/python-garminconnect) library (v0.3.2):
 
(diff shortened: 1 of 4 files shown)


victory-c:

victory-c:
Follow-up in the same confirmation-gate style for workouts: #334 (`delete_workout` / `delete_workouts`).

